PR #1, opened 2016-11-06
Mixify pairs based on the week

ericworkman:
@mbramson This deterministically generates pairs for a given week. Front page shows the pairs, given there are any users in the system.


mbramson:
Awesome! LGTM

PR #14, opened 2016-11-10
Add active column to Users index

### LINKED ISSUE #9: Add Active column to the /users index page
To see whether a user is active, you currently need to be look at a specific user. It would be nice if we could see which users were active at a glance when looking at all users. Bonus points if you could also toggle them active/inactive from the list as well.

joelbyler:
@mbramson looking for feedback on the `active_text` function in the User view

Closes #9


ericworkman:
@joelbyler You have a failing test

joelbyler:
@ericworkman we fixed it, how does everything look now?

mbramson:
LGTM!

PR #15, opened 2016-11-10
Fix conn build warning in page controller tests

ericworkman:
@mbramson The tests were warning about using build_conn instead of conn. This removes that warning.


mbramson (approved):
LGTM!

PR #16, opened 2016-11-12
Create Pair model and add feature to repairify

### LINKED ISSUE #3: Add Pair Model
Add a pair model that captures a given pairing session. This will be useful for a number of things down the line including histories and preventing users from being paired with users they paired with the week previous, 

At a minimum it would have the following fields:
* Date of Pair
* Many users
  * 2 or more (I wouldn't think there should be any limit)

Some other fields to think about that probably don't make sense yet:
* Subject of the pairing session
* Boolean for whether the pairing session actually occurred

ericworkman:
@mbramson Creates a pair model with a 3-part composite key. The front page now stores and retrieves these pairs instead of recalculating the pairs.  Added a button to repairify active users. Added routes to show the pairs of a specific year and week (and to repairify, probably want to change that in the future). Made the pairification consistent with a better query to fetch users.

There's some refactoring we can do once we want to add projects to the pairs. We might want to move the indexing to Pairer, refactor the template, move the Pair query to the model, etc. Open to suggestions or further issues.

Closes #3

### DIFF
--- a/web/controllers/page_controller.ex
+++ b/web/controllers/page_controller.ex
@@ -1,8 +1,9 @@
 defmodule Pairmotron.PageController do
   use Pairmotron.Web, :controller
 
-  alias Pairmotron.User
   alias Pairmotron.Pair
+  alias Pairmotron.User
+  alias Pairmotron.UserPair
   alias Pairmotron.Mixer
   alias Pairmotron.Pairer
 
@@ -48,9 +49,7 @@ defmodule Pairmotron.PageController do
 
   defp fetch_users_from_pairs(pairs) do
     pairs
-      |> Repo.preload([:user])
-      |> Enum.map(fn(p) -> p.user end)
-      |> Pairer.generate_pairs
+      |> Repo.preload([:users])
   end
 
   defp generate_pairs(year, week) do
@@ -66,7 +65,8 @@ defmodule Pairmotron.PageController do
   end
 
   defp make_pair(users, index, year, week) do
+    pair = Repo.insert! %Pair{year: year, week: week, pair_group: index}
     users
-      |> Enum.map(fn(user) -> %Pair{year: year, week: week, user_id: user.id, pair_group: index} end)
+      |> Enum.map(fn(user) -> %UserPair{pair_id: pair.id, user_id: user.id} end)
   end
 end
--- a/web/controllers/page_controller.ex
+++ b/web/controllers/page_controller.ex
@@ -30,12 +30,13 @@ defmodule Pairmotron.PageController do
   end
 
   defp fetch_or_gen(year, week) do
-    case fetch_pairs(year, week) |> fetch_user_pairs do
+    case fetch_pairs(year, week) do
       [] ->
         generate_pairs(year, week)
-        fetch_pairs(year, week) |> fetch_user_pairs
+        fetch_pairs(year, week)
       pairs -> pairs
     end
+      |> fetch_users_from_pairs
   end
 
   defp fetch_pairs(year, week) do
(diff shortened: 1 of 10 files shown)


web/models/pair.ex
```diff
@@ -0,0 +1,21 @@
...
+
+    timestamps()
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:year, :week, :pair_group, :user_id])
```
Serneum: Any opposition to making the set of columns a variable?

``` elixir
@required_fields ~w(year week pair_group user_id)
@optional_fields ~w()
```

ericworkman: That I can do

web/controllers/page_controller.ex
```diff
@@ -2,14 +2,71 @@ defmodule Pairmotron.PageController do
...
+        generate_pairs(year, week)
+        fetch_pairs(year, week)
+      pairs -> pairs
+    end
+      |> fetch_users_from_pairs
+  end
+
+  defp fetch_pairs(year, week) do
+    Pair
+      |> where(year: ^year, week: ^week)
```
mbramson, line 45: Why did you have to pin these?

web/models/pair.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Pairmotron.Pair do
+  use Pairmotron.Web, :model
+
+  schema "pairs" do
+    field :year, :integer
+    field :week, :integer
+    field :pair_group, :integer
+    belongs_to :user, Pairmotron.User
```
mbramson: Why did you choose to have one pair per user rather than a pair with a has_many relationship to users? That way if you had 10 users, you would have 5 pairs (and 5 db rows).

web/templates/page/index.html.eex
```diff
@@ -1,4 +1,4 @@
-<h2>Pairs for <%= Timex.today %></h2>
+<h2>Pairs for <%= @year %> week <%= @week %></h2>
```
mbramson: how about:
`<h2>Pairs for week <%= @week %> of <%= @year %></h2>` ?

I think it sounds a bit less awkward.

mbramson (commented):
Awesome! One question about the data model, but otherwise works great!

web/controllers/page_controller.ex
ericworkman: Variables in Ecto queries have to be pinned. Ecto will raise an error if you don't

```
** (Ecto.Query.CompileError) variable `year` is not a valid query expression. Variables need to be explicitly interpolated in queries with ^
```

web/models/pair.ex
ericworkman: Yep, that makes sense, especially for when we extend Pair for session information. Will do this.

web/templates/page/index.html.eex
ericworkman: Sure thing

web/controllers/page_controller.ex
mbramson: Huh! Didn't know that!

ericworkman:
@mbramson There we go. Created UserPair as a many-to-many. I was considering moving pair_group to UserPair, but I held off for now. It works fine where it is right now. Made the associations `:delete_all`, which means if a user is deleted or a pair is deleted (repairing), then the many-to-many records are deleted too.

mbramson:
Do we need the pair_group at all? I'm not sure what we get from having it. Why does it matter which number a pair is within a certain week? `:delete_all` works from now, but if we ever do any kind of statistics, we'd probably want to revisit that.

ericworkman:
I was using it to keep all the pairs in order. Its not so much needed, actually. Want me to remove it?

mbramson:
I'd think so. Couldn't we just order by the pair id? I feel like having it just adds another variable that we don't really need.

ericworkman:
Yep, that's what I'm doing now.

ericworkman:
@mbramson Removed pair_group. Also switched to using changesets for pairs and pairusers, cause its better.

mbramson:
Solid! LGTM

PR #20, opened 2016-11-15
Add whitespace between Back and Edit links

### LINKED ISSUE #18: Missing whitespace between Edit and Back links when viewing a project
image

Serneum:
Closes #18

### DIFF
--- a/web/templates/user/show.html.eex
+++ b/web/templates/user/show.html.eex
@@ -19,5 +19,5 @@
 
 </ul>
 
-<%= link "Edit", to: user_path(@conn, :edit, @user) %>
+<%= link "Edit", to: user_path(@conn, :edit, @user), class: "padded-link" %>
 <%= link "Back", to: user_path(@conn, :index) %>
--- a/web/static/css/app.css
+++ b/web/static/css/app.css
@@ -1,1 +1,4 @@
 /* This file is for your main application css. */
+a.padded-link {
+  padding-right: 5px;
+}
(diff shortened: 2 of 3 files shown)


Serneum:
Note: `&nbsp;` ends up adding two spaces for some reason. Adding whitespace at the end of the first link is ignored. Best solution will be one of the following:
1. Put both links on the same link in code
2. Add CSS to add padding/marging

web/templates/user/show.html.eex
```diff
@@ -19,5 +19,5 @@
 
 </ul>
 
-<%= link "Edit", to: user_path(@conn, :edit, @user) %>
+<%= link "Edit", to: user_path(@conn, :edit, @user) %>&nbsp;
```
ericworkman: You could

```
<%= link "Edit", to: user_path(@conn, :edit, @user), class: "btn btn-default btn-xs" %>
```

Serneum: Do we want links that look like buttons instead of links?

ericworkman: It was mostly an example of how to add class to the buttons. I don't care either way, but the buttons in bootstrap deal with spacing pretty well on their own.

mbramson: +1 for solving this with css rather than throwing a &nbsp in there.

Serneum:
This is ready for review

mbramson (approved):
LGTM!

PR #27, opened 2016-11-16
Add functionality for activating/deactivating users in the users list

### LINKED ISSUE #25: Allow users to be toggled active and inactive from the /users index route
It would be great to be able to toggle users active and inactive directly from the /users table, so that we don't have to manually update via the users specific show route.

Serneum:
Closes #25

### DIFF
--- a/test/controllers/user_controller_test.exs
+++ b/test/controllers/user_controller_test.exs
@@ -63,4 +63,28 @@ defmodule Pairmotron.UserControllerTest do
     assert redirected_to(conn) == user_path(conn, :index)
     refute Repo.get(User, user.id)
   end
+
+  test "activates chosen resource", %{conn: conn} do
+    user = Repo.insert! %User{name: "foo", email: "bar", active: false}
+    conn = get conn, user_path(conn, :index)
+    assert html_response(conn, 200) =~ "Inactive"
+
+    conn = put conn, user_activate_path(conn, :activate, user)
+    assert redirected_to(conn) == user_path(conn, :index)
+
+    conn = get conn, user_path(conn, :index)
+    assert html_response(conn, 200) =~ "Active"
+  end
+
+  test "inactivates chosen resource", %{conn: conn} do
+    user = Repo.insert! %User{name: "foo", email: "bar", active: true}
+    conn = get conn, user_path(conn, :index)
+    assert html_response(conn, 200) =~ "Active"
+
+    conn = put conn, user_deactivate_path(conn, :deactivate, user)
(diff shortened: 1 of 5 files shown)


test/controllers/user_controller_test.exs
```diff
@@ -63,4 +63,28 @@ defmodule Pairmotron.UserControllerTest do
...
+    user = Repo.insert! %User{name: "foo", email: "bar", active: true}
+    conn = get conn, user_path(conn, :index)
+    assert html_response(conn, 200) =~ "Active"
+
+    conn = put conn, user_deactivate_path(conn, :deactivate, user)
+    assert redirected_to(conn) == user_path(conn, :index)
+
+    conn = get conn, user_path(conn, :index)
+    assert html_response(conn, 200) =~ "Inactive"
+  end
```
mbramson, line 89: I think that these tests are testing too many things at once. You've even left whitespace where these tests should probably be broken out.

web/controllers/user_controller.ex
```diff
@@ -38,28 +38,49 @@ defmodule Pairmotron.UserController do
...
+    |> redirect(to: user_path(conn, :index))
+  end
+
+  def activate(conn, %{"user_id" => id}) do
+    update_user(conn, id, %{active: true}, :index)
+  end
+
+  def deactivate(conn, %{"user_id" => id}) do
+    update_user(conn, id, %{active: false}, :index)
+  end
```
mbramson: We should definitely explore ways to accomplish this without adding new non-RESTful controller methods, as discussed.

mbramson (changes requested):

Serneum:
@mbramson Commit one isn't RESTful, but cleaner in my opinion. Commit two is the more RESTful but I hate it. Tell me which you'd prefer.

web/router.ex
```diff
@@ -17,7 +17,10 @@ defmodule Pairmotron.Router do
     pipe_through :browser # Use the default browser stack
 
     get "/", PageController, :index
-    resources "/users", UserController
+    resources "/users", UserController do
+      put "/activate", UserController, :activate, as: :activate
+      put "/deactivate", UserController, :deactivate, as: :deactivate
```
daveshah, line 22: This doesn't seem to follow RESTful type behavior. Why are 'activate' and 'deactivate' their own routes here?

mbramson:
Closing in favor of https://github.com/ElixirCLE/pairmotron/issues/85, since the /users route no longer exists. That issue is a great opportunity to introduce react into this project.

Issue #24, opened 2016-11-15
Disable the ability to generate pairs for weeks that are not the current week


mbramson:
Currently, if you go to a week that is in the past, new pairs are automatically generated for that week. We should disable this functionality for past weeks. Maybe a screen that says "There were no pairs for this week".

Note that this is different than issue #23, as this issue is about disabling the automatic generation of new pairs when you go to a previous week where there were no existing pairs.

mbramson:
per a discussion with @ericworkman it doesn't really make sense to calculate pairs for future weeks either. This means that we don't have to implement past_week logic and can simply compare the Timex.iso_week output and not recalculate if either the year or week or different.

mbramson:
It looks like this unintentionally solved by #29

PR #29, opened 2016-11-18
23: Disable pairing for non current weeks

### LINKED ISSUE #23: Disable recalculation of pairs for weeks that are not the current week
Pairs for past weeks can be recalculated by clicking Repairify for routes in the past. This shouldn't be possible, as theoretically a past week's pairs already happened and should be set in stone. This will be useful if we eventually want to generate statistics on pairing over time.

Note that this issue is different from #24, as this is about disabling the Repairify button for weeks in the past where there were already existing pairs.

mbramson:
Reconfigure the show route to check if it is the current week before it
calculates a new set of pairs. If it is not the current week, it does
not calculate pairs.

Add Pairmotron.Time for determining if it is the current week.

Note that you can still see who paired for weeks that aren't the current week, but the automatic generation of pairs for those weeks does not occur if there are no current pairs for those weeks.

Currently, the Repairify button still works for all weeks, but that will be addressed in #24 

Closes #23

### DIFF
--- a/lib/time.ex
+++ b/lib/time.ex
@@ -1,0 +1,13 @@
+defmodule Pairmotron.Time do
+
+  @doc """
+  Returns true if the year and week passed in are the same week as the current
+  date. Weeks start on Monday. The first week of the year is defined as the first
+  week that contains a Thursday.
+  """
+  def same_week?(year, week, current_date) do
+    {curr_year, curr_week} = Timex.iso_week(current_date)
+    curr_year == year && curr_week == week
+  end
+
+end
--- a/test/controllers/page_controller_test.exs
+++ b/test/controllers/page_controller_test.exs
@@ -22,23 +22,35 @@ defmodule Pairmotron.PageControllerTest do
     setup do
       {:ok, usera} = Pairmotron.Repo.insert(%Pairmotron.User{name: "usera", email: "a", active: true})
       {:ok, userb} = Pairmotron.Repo.insert(%Pairmotron.User{name: "userb", email: "b", active: true})
-      conn = get build_conn, page_path(build_conn, :show, 2016, 40)
+      {year, week} = Timex.iso_week(Timex.today)
+      conn = get build_conn, page_path(build_conn, :show, year, week)
       {:ok, [conn: conn, usera: usera, userb: userb]}
     end
 
-    test "saves and stores the pairs", %{conn: conn, usera: usera, userb: userb} do
+    test "displayse the users that were paired", %{conn: conn, usera: usera, userb: userb} do
       assert html_response(conn, 200) =~ usera.name
       assert html_response(conn, 200) =~ userb.name
+    end
+
+    test "does not display a user added after first pairing", %{conn: conn} do
       {:ok, userc} = Pairmotron.Repo.insert(%Pairmotron.User{name: "userc", email: "c", active: true})
-      conn = get conn, page_path(conn, :show, 2016, 40)
+      {year, week} = Timex.iso_week(Timex.today)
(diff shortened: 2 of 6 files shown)


lib/time.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Pairmotron.Time do
```
ericworkman: I think `Calendar` might be a better module name, since it won't conflict with Elixir's Time module like this will. And its solely dealing with weeks and dates, not times.

test/controllers/page_controller_test.exs
```diff
@@ -22,23 +22,35 @@ defmodule Pairmotron.PageControllerTest do
     setup do
       {:ok, usera} = Pairmotron.Repo.insert(%Pairmotron.User{name: "usera", email: "a", active: true})
       {:ok, userb} = Pairmotron.Repo.insert(%Pairmotron.User{name: "userb", email: "b", active: true})
-      conn = get build_conn, page_path(build_conn, :show, 2016, 40)
+      {year, week} = Timex.iso_week(Timex.today)
```
ericworkman, line 25: 👍 for removing my hard coded date

test/controllers/page_controller_test.exs
```diff
@@ -22,23 +22,35 @@ defmodule Pairmotron.PageControllerTest do
...
+      conn = get build_conn, page_path(build_conn, :show, year, week)
       {:ok, [conn: conn, usera: usera, userb: userb]}
     end
 
-    test "saves and stores the pairs", %{conn: conn, usera: usera, userb: userb} do
+    test "displayse the users that were paired", %{conn: conn, usera: usera, userb: userb} do
```
ericworkman: `displays`

lib/time.ex
```diff
@@ -0,0 +1,13 @@
...
+
+  @doc """
+  Returns true if the year and week passed in are the same week as the current
+  date. Weeks start on Monday. The first week of the year is defined as the first
+  week that contains a Thursday.
+  """
+  def same_week?(year, week, current_date) do
```
ericworkman: Would it be worthwhile to change signature to `same_week?/2`? Mostly I mean change the arguments to `{year, week}, current_date` here and clean up all the calls?

ericworkman (changes requested):
Pretty sweet! Just a couple of things to look at

lib/time.ex
mbramson: Good. Call.

test/controllers/page_controller_test.exs
mbramson: Glad you like this, because I felt slightly dirty doing this. Not a huge issue, but this will fail at midnight on Sundays if you run the tests at just the right time.

lib/time.ex
mbramson: Did some thinking on this. It doesn't make a huge difference really. Making a function have less parameters is always a good thing in a vacuum, but it makes calling it look a little more complicated (not a strong argument there). The thing I like most about making it two parameters, is that it makes both parameters more similar to each other. They're still not identical, since the second parameter is actually a date object, but they are more similar.

test/controllers/page_controller_test.exs
ericworkman: Yes, it would, but I'd like to see you make that happen ;)

mbramson:
@ericworkman updated the PR per your concerns. I left the same_week? parameter change in its own commit, because I'm not sure that I like it now that I've typed it out. I think that the ternary signature is simpler and more readable. What do you think?

ericworkman:
Looks good, but I agree. It looks better and works better with what we have as /3 not /2.

PR #31, opened 2016-11-22
Add user authentication

### LINKED ISSUE #5: Add User Authentication
Doesn't have to be super fancy or secure, but it would be nice to get something in place so that a current user can be put in session for other functionality down the line such as groups, admin, etc.

mbramson:
Add user authentication.

Add the session and registration paths.

Add ex-machina for testing and comeonin for password hashing.

Any accounts that currently exist before the migration in this PR will be assigned a password of "password" (which they can change by editing their user).

Closes #5


mbramson:
So I think that I'm going to lead Authentication tokens (like https://github.com/ueberauth/guardian) to a separate PR. It's a great exercise, and this PR is already fairly massive as it is.

web/models/user.ex
```diff
@@ -6,19 +6,47 @@ defmodule Pairmotron.User do
...
+  def registration_changeset(struct, params \\ %{}) do
+    struct
     |> cast(params, @required_params, @optional_params)
-    |> validate_required([:name, :email])
+    |> unique_constraint(:email)
+    |> validate_length(:password, min: @minimum_password_length)
+    |> validate_length(:password_confirmation, min: @minimum_password_length)
+    |> validate_confirmation(:password)
+    |> generate_password
+  end
```
daveshah, line 46: Could these two (`changeset` and `registration_changeset` )use a common private function to reduce the duplication?

web/router.ex
```diff
@@ -9,14 +9,29 @@ defmodule Pairmotron.Router do
...
     plug :accepts, ["json"]
   end
 
   scope "/", Pairmotron do
-    pipe_through :browser # Use the default browser stack
+    pipe_through :browser
+
+    get "/registration", RegistrationController, :new
+    post "/registration", RegistrationController, :create
+
```
daveshah: could we replace this with `resources "registration", RegistrationController, only: [:new, :create]` ?

web/models/user.ex
mbramson: I was thinking about that actually. The answer is yes.

web/router.ex
mbramson: Definitely.

web/models/user.ex
```diff
@@ -6,19 +6,47 @@ defmodule Pairmotron.User do
...
+
+  def generate_password(changeset) do
```
daveshah: should this be private?

mbramson: Yep. Before, it was actually called in the user and registration create and new routes, which necessitated it being public. Then realized it should probably just be part of the changeset itself.

web/models/user.ex
```diff
@@ -6,19 +6,47 @@ defmodule Pairmotron.User do
...
+    case changeset do
+      %Ecto.Changeset{valid?: true, changes: %{password: pass}} ->
+        put_change(changeset, :password_hash, Comeonin.Bcrypt.hashpwsalt(pass))
```
daveshah: It feels weird that we're sprinkling this dependency around (also [here](https://github.com/ElixirCLE/pairmotron/pull/31/files#diff-e8a7c764b6ea7aeebe12d2ecd2313a5aR36) ) I'm wondering if Password type stuff should be extracted into its own module - putting an interface in front of this dependency `¯\_(ツ)_/¯`

mbramson: That's a good call-out. My thought process in not making it its own module was admittedly partly due to the various projects/tutorials I looked at implementing this in the user model. It was also placed here partly because it modifies a changeset (and only a User changeset). Another (not as great) argument for it being here is that it really isn't a ton of logic.

mbramson: So now that there is a common_changeset method that only calls generate_password once I think the argument for keeping here is slightly stronger

mbramson:
@daveshah I still need to adapt the tests that test generate_password directly to just test the changesets themselves, but otherwise I think I've addressed your concerns.

edit: Looking at the tests that I have so far, I think the generate_password functionality is covered by the SessionController tests.

web/models/user.ex
mbramson: This has been done. Let me know what you think of the implementation.

mbramson:
Disabled the ability for users to set new passwords for users that are not themselves per @ericworkman 's suggestions. I'm leaving the password and password_confirmation fields on the user edit form universally for now, they just cannot be used when looking at a different user (an error is thrown).

mbramson:
Updated the database migration so that it assigns a password of "password" to all users that exist before the migration.

web/controllers/user_controller.ex
```diff
@@ -62,4 +72,11 @@ defmodule Pairmotron.UserController do
     |> put_flash(:info, "User deleted successfully.")
     |> redirect(to: user_path(conn, :index))
   end
+
+  defp current_assigned_user_id(conn) do
```
ericworkman, line 76: Near duplicate of https://github.com/ElixirCLE/pairmotron/pull/31/files#diff-c9ec1fc6e91a7bd967c8f8506027d1f9R27.

See the note for https://github.com/ElixirCLE/pairmotron/pull/31/files#diff-7dc24a26994057cb34cfbdb463c15b3fR44

web/controllers/user_controller.ex
```diff
@@ -41,6 +41,16 @@ defmodule Pairmotron.UserController do
     user = Repo.get!(User, id)
     changeset = User.changeset(user, user_params)
 
+    if current_assigned_user_id(conn) != id && changeset.changes[:password] do
```
ericworkman, line 44: Since user is from `Repo.get!`, it should be safe to use `user.id` instead of `id` here. Then you don't have the `to_string` cast in `current_assigned_user_id`.

web/templates/layout/app.html.eex
```diff
@@ -20,6 +20,14 @@
             <li><%= link "Pairs", to: page_path(@conn, :index) %></li>
             <li><%= link "Users", to: user_path(@conn, :index) %></li>
             <li><%= link "Projects", to: project_path(@conn, :index) %></li>
+            <li>
+              <%= case @conn.assigns[:current_user] do %>
+                <% nil -> %>
+                  <%= link "Login", class: "btn btn-success", role: "button", to: session_path(@conn, :new) %>
+                <% current_user -> %>
```
ericworkman: Isn't used, so it should be `_current_user`

ericworkman (approved):
Looks good. Fix these minor things in this PR or in others.

web/templates/layout/app.html.eex
mbramson: Great call.

web/controllers/user_controller.ex
mbramson: I am going to leave this for a separate PR, but you are correct. This should be refactored.

PR #33, opened 2016-12-03
Allow users to be toggled active and inactive from the /users index route #25

### LINKED ISSUE #25: Allow users to be toggled active and inactive from the /users index route

daveshah:
Still WIP.

Closes #25


web/controllers/user_controller.ex
```diff
@@ -5,7 +5,9 @@ defmodule Pairmotron.UserController do
 
   def index(conn, _params) do
     users = Repo.all(User)
-    render(conn, "index.html", users: users)
+    changesets = users |> Enum.map &User.changeset/1
+    user_changesets = Enum.zip(users, changesets)
+    render(conn, "index.html", user_changesets: user_changesets)
   end
```
daveshah, line 11: @Serneum & @mbramson - this is what I had in mind. If you're view requires a changeset, simply supply one 
`¯\_(ツ)_/¯`

web/templates/user/index.html.eex
```diff
@@ -1,30 +1,9 @@
...
-
-      <td class="text-right">
-        <%= link "Show", to: user_path(@conn, :show, user), class: "btn btn-default btn-xs" %>
-        <%= link "Edit", to: user_path(@conn, :edit, user), class: "btn btn-default btn-xs" %>
-        <%= link "Delete", to: user_path(@conn, :delete, user), method: :delete, data: [confirm: "Are you sure?"], class: "btn btn-danger btn-xs" %>
-      </td>
-    </tr>
+<div class="container">
+<%= for {user, changeset} <- @user_changesets do %>
+    <%= render "form.html", changeset: changeset, action: user_path(@conn, :update, user), conn: @conn %>
```
daveshah: @mbramson @Serneum once the changeset is available, we can then have a form/per user listed. I reused the form that was there to keep it simple but we could create another (with mo' better styling) to get the effect that's desired.

daveshah:
@mbramson @Serneum this approach prevents created routes for actions, still keeping things restful but providing what's needed to update users from index.

daveshah:
Updated and added `short_form.html.eex`

mbramson:
Closing in favor of https://github.com/ElixirCLE/pairmotron/issues/85, since the /users route no longer exists. That issue is a great opportunity to introduce react into this project.

PR #40, opened 2016-12-08
39: Add guardian token to authentication

### LINKED ISSUE #39: Add token-based authentication
Add token-based authentication. https://github.com/ueberauth/guardian is a prime candidate for this.

mbramson:
Adds token-based authentication.

Simplifies the Authenticate Plug because the User is retrieved from the database directly by Guardian whenever authentication occurs. This speeds up tests for authenticated routes slightly.

Closes #39

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -22,6 +22,21 @@ config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
+config :guardian, Guardian,
+  allowed_algos: ["ES512"],
+  verify_module: Guardian.JWT,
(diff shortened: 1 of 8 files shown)


ericworkman:
@mbramson I've added the appropriate config to Heroku for this. The key was simply generated from `mix phoenix.gen.secret`. We can spend the time to generate a better key using params or pem files, but this should be fine. I don't think there's any need to add the generation method to the readme, unless you really want to.

mbramson:
I don't think that there is any need either. LGTM.

ericworkman (approved):

PR #41, opened 2016-12-10
Make description and url optional in a project

### LINKED ISSUE #38: Make the url and description not required for projects
Urls should not be a required field for projects. Not all potential projects to pair on are code related or on the internet. It could be "put together that table in the corner", for example.

Descriptions also should not always be required. Sometimes a project is simple enough that the title is sufficient.

ericworkman:
Closes #38


mbramson (approved):
Solid. LGTM!

PR #46, opened 2016-12-14
Fix guardian

ericworkman:
@mbramson


mbramson (approved):
LGTM! If we ever need to we can always switch the prod algorithm to something else other than HS512.

PR #47, opened 2016-12-18
32 - Add pair retrospective

### LINKED ISSUE #32: Add the ability to state what you have paired on
This could either just be a string field on the Pair model, or perhaps a link to a specific project. I think that it should at the very least have a string field so that we gain some metrics on what users have been pairing on.

This would necessitate a new "post-pair" UI. Perhaps this would be accessible from the main /pairs route?

You should only be able to post about a pair that you actually participated in.

mbramson:
Adds pair retrospectives which allow users to document their pairing sessions.

Closes #32 

Cards to make:
- [x] Make it so that repairifying doesn't delete retrospectives. This probably means implementing some logic so that repairifying doesn't actually delete pair sessions if the resulting new pair sessions would be identical.
- [x] Refactor retro :create authorization restrictions to use Canary, not just the four routes that have id's associated. :new does not need it as the user_id is derived from within the controller.
- [x] Address the deletion of projects when retrospectives rely on them. This currently worked by nilifying the project_id on the affected retrospectives, but it might be nice to at least warn the user when this is going to occur.
- [x] You can now create a retrospective for a pair that you are not in. Which is bad.
- [x] Pass the projects for the new and edit routes of the pair_retro_controller rather than assigning to the conn.


lib/abilities.ex
```diff
@@ -0,0 +1,8 @@
+defimpl Canada.Can, for: Pairmotron.User do
+  alias Pairmotron.{PairRetro, User}
+
+  def can?(%User{id: user_id}, action, %PairRetro{user_id: user_id})
+    when action in [:edit, :update, :show, :delete], do: true
+
+  def can?(%User{}, _, _), do: false
+end
```
mbramson, line 8: The thought here is that canary will only be used on controller actions for now explicitly. If you use load_and_authorize_resource on something other than the above actions on the `PairRetroController`, you will have to whitelist it here.

Also of note, I am only using canary to authorize resources that can only be accessed by the associated user. At this time, this is only `PairRetro`s. This removed a ton of code from the pair retro controller actions, and made things much cleaner. It also removed a lot of duplicate code to popular and well maintained abstractions (`canada` and `canary`). The fact that they are well maintained and some of the go to solutions to authorizing specific resources mitigates the risk of adding these dependencies. They are also very lightweight compared to other solutions.

test/controllers/pair_retro_controller_test.exs
```diff
@@ -0,0 +1,179 @@
...
+
+    test "does not create retro with a pair_date before the pair's week", %{conn: conn, logged_in_user: user} do
+      pair = create_pair([user], 2016, 1)
+      attrs = Map.merge(@valid_attrs, %{pair_date: ~D(1999-10-20),
+                                        pair_id: Integer.to_string(pair.id),
+                                        user_id: Integer.to_string(user.id)})
+      conn = post conn, pair_retro_path(conn, :create), pair_retro: attrs
+      assert html_response(conn, 200) =~ "New retrospective"
+      assert html_response(conn, 200) =~ "cannot be before the week of the pair"
+    end
```
mbramson, line 76: The test around the behavior for creating a retrospective for a date that occurs in the future is contained in the `PairRetroTest` model tests, as all of its behaviors is contained in the changeset validation itself. This test exists in here, as the decision was made to pass in the date of the pair to the changeset. See the comment around the changeset for more details.

test/lib/calendar_test.exs
```diff
@@ -20,4 +20,29 @@ defmodule Pairmotron.CalendarTest do
...
+    test "past week in past year returns true" do
+      assert Calendar.past_week?(2015, 50, ~D(2016-02-01))
+    end
+  end
+
+  describe "first_date_of_week" do
+    test "returns the date of the first day of the given week" do
+      assert Calendar.first_date_of_week(2016, 5) == ~D(2016-02-01)
+      assert Calendar.first_date_of_week(2015, 53) == ~D(2015-12-28)
+    end
```
mbramson: Can anyone think of any other dates that should be tested?

web/models/pair_retro.ex
```diff
@@ -0,0 +1,70 @@
...
+  """
+  def changeset(struct, params \\ %{}, pair_start_date) do
+    struct
+    |> cast(params, @required_fields, @optional_fields)
+    |> foreign_key_constraint(:user_id)
+    |> foreign_key_constraint(:pair_id)
+    |> foreign_key_constraint(:project_id)
+    |> validate_field_is_not_before_date(:pair_date, pair_start_date)
+    |> validate_field_is_not_in_future(:pair_date)
+  end
```
mbramson, line 35: So the decision was made based on conversations with @ericworkman and with denizens of the Ecto room of the elixir-lang slack to pass in an additional parameter to the changeset (`pair_start_date`) rather than the alternative of making a database call from within the `validate_field_is_not_before_date` changeset validator. I'm on the fence about it.

On the side of passing the `pair_start_date` into the changeset, it isolated all of the database interaction to the controller. This injects the pair_start_date dependency into the changeset function, thereby confining the database lookup side effect to the controller.

There is an argument for not passing the pair_start_date parameter into the changeset at all, and instead invoking `Repo.get(Pairmotron.Pair, pair_id)` from within the `validate_field_is_not_before_date` function to retrieve the `pair_start_date`. The database call would be called through `Ecto.Changeset.validate_change`, which only invokes the logic if the given field `:pair_date` actually changes. This eliminates a database select when the field isn't actually changed. This logic (eliminating unnecessary database calls) could be duplicated in the controller, but it would be duplicated functionality, and would add complexity to the controller. The downside of this approach is making the `validate_field_is_not_before_date` less abstract, and also impure.

test/lib/calendar_test.exs
ericworkman: `Calendar.first_date_of_week(2004, 1)` and `Pairmotron.Calendar.first_date_of_week(2004, 9)` and `Pairmotron.Calendar.first_date_of_week(2004, 10)`

ericworkman: Looks like it is doing things mostly correctly for ISO weeks, but I did find this:

```
>> Pairmotron.Calendar.first_date_of_week(2004, 1)
** (FunctionClauseError) no function clause matching in Timex.Helpers.iso_day_to_date_tuple/2
    (timex) lib/timex/helpers.ex:6: Timex.Helpers.iso_day_to_date_tuple(2004, -2)
    (timex) lib/timex.ex:612: Timex.from_iso_triplet/1
```

mbramson: As discussed, this was a bug in Timex itself. This should be fixed by this PR: https://github.com/bitwalker/timex/pull/257

I'll add some additional tests that fail until that timex PR is merged.

mbramson:
Updated Timex so that it points to the timex commit hash of the fix to the bug that @ericworkman discovered. Added a test that failed before the commit hash was added and passed afterwards. I will update the dependency to point at a version as soon Timex is updated on hex.pm, although that might be after this has been merged.

ericworkman:
Looks good 👍

PR #49, opened 2016-12-27
Fix error when logging in with unknown email

mbramson:
Logging in with an email that doesn't exist was resulting in an
unhandled exception. The test case around this did not catch it because
the user factory generates a nil password, which is caught by a function
head that handled a nil password correctly.

### DIFF
--- a/test/controllers/session_controller_test.exs
+++ b/test/controllers/session_controller_test.exs
@@ -32,8 +32,7 @@ defmodule Pairmotron.SessionControllerTest do
   end
 
   test "logging in with an email that doesn't exist fails with error message", %{conn: conn} do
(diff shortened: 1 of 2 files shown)


test/controllers/session_controller_test.exs
```diff
@@ -32,8 +32,7 @@ defmodule Pairmotron.SessionControllerTest do
   end
 
   test "logging in with an email that doesn't exist fails with error message", %{conn: conn} do
-    %User{email: _user_email, password: user_pw} = insert(:user)
```
mbramson, line 35: Not sure why I thought I needed to insert a user at all for this test.

ericworkman (approved):

PR #51, opened 2017-01-03
Remove ability to delete other users

### LINKED ISSUE #50: Disable ability to delete other users
Users can delete other users right now. This should not be possible. You should be able to delete your own user but not someone else's. You should be able to use an out of the box `authorize_resource` plug from Canary to do this.

For now, you should still be able to edit other users. Once #35 has been addressed, all user edit and update actions should also be limited to the logged in user or an administrator.

mbramson:
Prevents users from deleting other users. Forces the user to authorize using canary for the delete action of the user controller, and also hides the delete button for all users in the user index that are not the current user.

Closes #50

### DIFF
--- a/web/controllers/user_controller.ex
+++ b/web/controllers/user_controller.ex
@@ -4,7 +4,7 @@ defmodule Pairmotron.UserController do
   alias Pairmotron.User
   import Pairmotron.ControllerHelpers
 
-  plug :load_and_authorize_resource, model: User, only: [:delete]
+  plug :load_and_authorize_resource, model: User, only: [:edit, :update, :delete]
 
   def index(conn, _params) do
     users = Repo.all(User)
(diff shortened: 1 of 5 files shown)


web/controllers/user_controller.ex
```diff
@@ -61,16 +64,16 @@ defmodule Pairmotron.UserController do
     end
   end
 
-  def delete(conn, %{"id" => id}) do
-    user = Repo.get!(User, id)
-
-    # Here we use delete! (with a bang) because we expect
-    # it to always work (and if it does not, it will raise).
-    Repo.delete!(user)
+  def delete(conn, %{"id" => _id}) do
```
ericworkman: Is the struct match here needed?

mbramson: Nope! great call!

mbramson:
So I expanded the scope of this a bit. I additionally restricted users from editing or updating other users. I also hid our emails from the show route, to prevent users from broadcasting their emails to the world.

web/templates/user/index.html.eex
```diff
@@ -19,7 +19,9 @@
       <td class="text-right">
         <%= link "Show", to: user_path(@conn, :show, user), class: "btn btn-default btn-xs" %>
         <%= link "Edit", to: user_path(@conn, :edit, user), class: "btn btn-default btn-xs" %>
```
ericworkman: Should you move the edit button into the if just below?

mbramson: Absolutely. Done.

ericworkman:
What happens when a user deletes themselves? They get logged out?

mbramson:
That's exactly what happens.

ericworkman:
LGTM

PR #52, opened 2017-01-03
Make login redirect to /pairs when user in session

### LINKED ISSUE #44: Make the route for login redirect to the main page if a user is already logged in
Currently, when a user visits the login page when they are already logged in, they can log in again. There is no indication that they are logged in when they do so. Nothing explicitly breaks when they do this, but they should be redirected to the main page when this occurs since they are in fact already logged in.

mbramson:
Currently, if the user is logged in, and visits the "/" route, they are greeted by the login screen, with no indication that they are actually logged in. This doesn't quite make sense, so in this case, the user is redirected to the /pairs route.

Closes #44


ericworkman (approved):
LGTM

PR #55, opened 2017-01-08
35: Add roles and admin functions

### LINKED ISSUE #35: Add Administrator Role
Add a role table. Each role should have a has_many relationship to users.

Add user and administrator roles. The first user should be made the administrator. This should be the user with the lowest id.

Administrators should be able to make other users administrators as well. Users should not be able to change the role of other users.

Some things to think about:
- If there are no users, and a user is registered/created, they should be made an administrator.
  - Thought about this, probably safer just to see an admin user/role or maybe just the role itself.
  - First user being an admin, might be convenient for someone hosting pairmotron, but might actually be confusing from a development standpoint for newer contributors down the line.
- In the migration that adds the administrator role, the user with the lowest id should be made an administrator.
  - After some thought, probably not a good idea.
- Only administrators should be allowed to delete users.
- If an administrator is deleted and there are no more administrators, the user with the lowest id should be made an administrator.
  - After some thought, probably not a good idea.
  - There should be some sort of warning or confirmation when a user tries to delete the last administrator that informs them that they are doing so and that a new user will be selected to be the only administrator.
  - A warning might still be nice, but I think this is outside of the scope of the issue.

mbramson:
Adds the roles table. Users can have Roles, which describe what they are able to do in the app. As a part of this PR, Roles have a name and a boolean value indicating whether they are admins or not.

Currently admins' only special abilities are that they can perform all of the functions for users and retrospectives that are currently only modifiable when they are associated with the logged in user.

The proposed logic to make the first user an admin by default, as specified in the original issue, has been decided against. Although it might make someone first using the app's life a bit easier, it might also be unintuitive from a development standpoint. A better option to make sure that an admin is setup initially would be to either seed an admin user and an admin role, or only the role. Anyone have thoughts?

- [x] Figure out what to seed.

I was considering adding the ability for admin's to set the roles of other users, but ultimately decided against it for this PR. The implementation details are a bit more nuanced than they first seem, as there are a few ways to do it with varying levels of db performance and complexity. Ultimately, this PR works as is, but requires some manual db intervention to setup roles and assign users roles.

- [ ] Make an issue to add role selection to the UI.

closes #35

### DIFF
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -1,11 +1,18 @@
 defimpl Canada.Can, for: Pairmotron.User do
-  alias Pairmotron.{PairRetro, User}
+  alias Pairmotron.{PairRetro, Role, User}
 
   def can?(%User{id: user_id}, action, %PairRetro{user_id: user_id})
     when action in [:edit, :update, :show, :delete], do: true
 
   def can?(%User{id: user_id}, action, %User{id: user_id})
     when action in [:edit, :update, :delete], do: true
+
+  def can?(%User{role_id: role_id}, _, _) when not is_nil(role_id) do
+    case Pairmotron.Repo.get(Role, role_id) do
+      nil -> false
+      role -> role.is_admin
+    end
+  end
 
   def can?(%User{}, _, _), do: false
 end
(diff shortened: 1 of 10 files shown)


ericworkman (approved):
LGTM

PR #59, opened 2017-01-09
Preload role for current_user

mbramson:
The role of the user is used enough for administrative purposes that it makes sense to preload it whenever we authorize the user.

This PR causes the role to be preloaded whenever assigning the current_user during authorization.

This PR also uses this in the users view to allow the edit and delete buttons to appear around any user if the current_user is an admin.

### DIFF
--- a/web/models/user.ex
+++ b/web/models/user.ex
@@ -19,7 +19,7 @@ defmodule Pairmotron.User do
   @minimum_password_length 8
 
   @required_params ~w(name email)
-  @optional_params ~w(password password_confirmation active)
+  @optional_params ~w(password password_confirmation active role_id)
 
   @doc """
   Builds a changeset based on the `struct` and `params`.
--- a/web/views/view_helper.ex
+++ b/web/views/view_helper.ex
@@ -1,0 +1,11 @@
+defmodule Pairmotron.ViewHelpers do
+  @doc """
+  Given a user, returns true if that user has a role and that role's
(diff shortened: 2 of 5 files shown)


web/models/user.ex
```diff
@@ -19,7 +19,7 @@ defmodule Pairmotron.User do
   @minimum_password_length 8
 
   @required_params ~w(name email)
-  @optional_params ~w(password password_confirmation active)
+  @optional_params ~w(password password_confirmation active role_id)
```
mbramson, line 22: Not actually used in this PR, but useful for development so that you can assign roles from iex.

web/views/view_helper.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Pairmotron.ViewHelpers do
+  @doc """
+  Given a user, returns true if that user has a role and that role's
+  is_admin property is true. Errors if the user's role is not loaded
+  and the user has a role. This is to prevent calling the database in
+  a view.
+  """
+  def is_admin?(user) do
+    user.role_id && user.role.is_admin
+  end
```
mbramson, line 10: So like the comment says, this errors if the user has a role and that role association is not preloaded. That's okay though, as it should always be preloaded at this point, and if it isn't, it should hard error rather than try to grab the role from the database (like the controller_helper version of this method does).

ericworkman (approved):
LGTM

PR #60, opened 2017-01-09
Fix current retro display bug in page :show

mbramson:
Currently, if the current user has submitted a retrospective for a pair, that retrospective is not assigned to the conn, and so isn't visible on the page :show route like it is for the page :index route. This PR fixes this.


mbramson:
Now, with tests!

ericworkman (approved):
LGTM

PR #62, opened 2017-01-22
Add groups

### LINKED ISSUE #10: Add Groups
Add groups so that users can be members of (multiple groups per user? Some more thinking here). Each user would then be paired within their own group.

There would need to be some sort of way to decide which group you were looking at in the page controller. The displayed pairs would filter accordingly.

Projects would also need to be associated with groups.

It would be nice if there was an invite mechanism for groups, so that the user that creates a group becomes an admin of that group, and is then the one that distributes invites to the group.

mbramson:
Adds groups models/migration and some basic UI around them.

### TODO:

- [x] Add group model/migration
- [x] Add group CRUD
- [x] Add resource specific authorization to groups based on owner
- [x] Add logic to set currently logged in user as owner for new groups
- [x] Add logic to add currently logged in user to the users of a new group
- [x] Add group link to toolbar
- [x] Display the owner for the group :show action

### New issues to create after this is complete. A number of these will be recategorized.

- [x] Allow groups to be private or public. Private groups would only be visible to members that are in them.
- [x] Add UI to transfer the ownership of a group
- [ ] Add support for group-specific roles, so that the owner isn't the only user that can add users or edit groups.
- [x] Allow users to request membership in a group.
- [x] Make Projects group-specific
- [x] Add ability for users to leave a group.
- [x] Add UI to allow owners of groups to remove users from a group
- [x] Add functionality for owners of groups to invite users to the groups
  - Discuss: Should this be its own model, or could this data belong on the UserGroup joining table?
  - Add UI to allow owners to select users that should receive invites to their group.
  - Add group_invitation model
  - Add group_invitation routes/controllers
  - Add group_invitation UI
- [x] Add dropdown to select Group for "/pairs"
- ~~Add ability to filter users by group on the users index~~
  - ~~Add filter for users in no groups~~
- [x] Display the user's groups on the user show action
- [ ] Display the group's users on the group show action (or perhaps a different controller?)
- [ ] Add landing page to redirect users to that aren't yet associated with any groups

Closes #10

### DIFF
--- a/web/models/group.ex
+++ b/web/models/group.ex
@@ -18,6 +18,7 @@ defmodule Pairmotron.Group do
   def changeset(struct, params \\ %{}) do
     struct
     |> cast(params, @required_fields, @optional_fields)
+    |> foreign_key_constraint(:owner_id)
   end
 
   @doc """
@@ -27,6 +28,7 @@ defmodule Pairmotron.Group do
   def changeset_for_create(struct, params \\ %{}, users) do
     struct
     |> cast(params, @required_fields, @optional_fields)
+    |> foreign_key_constraint(:owner_id)
     |> put_assoc(:users, users)
   end
 end
--- a/web/models/group.ex
+++ b/web/models/group.ex
@@ -4,17 +4,29 @@ defmodule Pairmotron.Group do
   schema "groups" do
     field :name, :string
     belongs_to :owner, Pairmotron.User
-    many_to_many :users, Pairmotron.User, join_through: "users_groups"
+    many_to_many :users, Pairmotron.User, join_through: Pairmotron.UserGroup
 
     timestamps()
   end
+
+  @required_fields ~w(name owner_id)
+  @optional_fields ~w()
 
   @doc """
   Builds a changeset based on the `struct` and `params`.
   """
   def changeset(struct, params \\ %{}) do
     struct
-    |> cast(params, [:name, :owner_id])
-    |> validate_required([:name])
+    |> cast(params, @required_fields, @optional_fields)
+  end
+
+  @doc """
+  Builds a changeset and sets the users association. Should be used
+  when creating a group so that the owner can be set as the only user.
+  """
+  def changeset_for_create(struct, params \\ %{}, users) do
+    struct
+    |> cast(params, @required_fields, @optional_fields)
+    |> put_assoc(:users, users)
   end
 end
--- a/web/models/group.ex
+++ b/web/models/group.ex
@@ -1,0 +1,21 @@
+defmodule Pairmotron.Group do
+  use Pairmotron.Web, :model
+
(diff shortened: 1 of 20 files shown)


ericworkman:
Can we break some of the checklist into cards and reduce the scope of this PR? I think this should deal with the models, CRUD, and the logic for group owner. 

We should have new cards for, in no particular order:

- Invitation to groups
- Display of groups
- Removal of users from groups
- Group specific pairing
- Other group roles

I think it may also be time to retire the /users portion from general-access to admin-only. We should pick up elm as we discussed, and use these new features as part of that.

mbramson:
Great call. I'll cut down the scope of this PR a bit to the point where groups can be created, and users can be added on the backend.

web/models/group.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Pairmotron.Group do
+  use Pairmotron.Web, :model
+
+  schema "groups" do
+    field :name, :string
+    belongs_to :owner, Pairmotron.User
+    many_to_many :users, Pairmotron.User, join_through: Pairmotron.UserGroup
```
mbramson, line 7: Fun fact: since this association is going to be used in a `Ecto.Changeset.put_assoc` call and the `users_groups` table has timestamps, we need to explicitly join through the Ecto schema. If we were just going to manually insert UserGroup records manually (which is the plan), we could just join through the column name only. This is what's done in the many_to_many association on the User model, since we don't need to use put_assoc or cast_assoc from that end.

web/templates/group/form.html.eex
```diff
@@ -0,0 +1,20 @@
+<%= form_for @changeset, @action, fn f -> %>
+  <%= if @changeset.action do %>
+    <div class="alert alert-danger">
+      <p>Oops, something went wrong! Please check the errors below.</p>
+    </div>
+  <% end %>
+
+  <%= hidden_input f, :owner_id, value: value_from_changeset(@changeset, :owner_id) %>
+  <%= error_tag f, :owner_id %>
```
mbramson, line 9: Basically shouldn't ever be used, but just in case something terribly wonky happens, it'd be nice for the foreign key constraint error (or others having to do with owner) to have somewhere to go that is visible to the user.

web/views/pair_retro_view.ex
```diff
@@ -1,10 +1,6 @@
 defmodule Pairmotron.PairRetroView do
   use Pairmotron.Web, :view
 
-  def value_from_changeset(changeset, value) do
-    elem(Ecto.Changeset.fetch_field(changeset, value), 1)
-  end
-
```
mbramson, line 7: Moved to ViewHelpers which is autoimported into Pairmotron.Web for views, since this is also used for groups.

ericworkman (approved):
LGTM

PR #63, opened 2017-01-24
Move the submit buttons to the right for regististration and login

ericworkman:
@mbramson I was annoyed that the "submit" button for registration and login were on the left instead of the right. Now they are on the right.


mbramson (approved):
Great call. LGTM.

PR #64, opened 2017-01-24
Remove Role and add User.is_admin

ericworkman:
@mbramson As we discussed, I've removed the Roles table and added the is_admin flag.


web/controllers/authenticate.ex
```diff
@@ -18,7 +17,7 @@ defmodule Pairmotron.Plug.Authenticate do
     assign_current_user(conn, current_user)
   end
   defp assign_current_user(conn, user = %User{}) do
-    assign(conn, :current_user, user |> Repo.preload(:role))
+    assign(conn, :current_user, user)
```
mbramson, line 20: 😀

web/controllers/controller_helpers.ex
```diff
@@ -50,12 +50,6 @@ defmodule Pairmotron.ControllerHelpers do
...
   """
   def is_admin?(user) do
-    cond do
-      is_nil(user.role_id) -> false
-      Ecto.assoc_loaded?(user.role) ->
-        user.role.is_admin
-      true ->
-        Pairmotron.Repo.get!(Pairmotron.Role, user.role_id).is_admin
-    end
+    user.is_admin
```
mbramson, line 53: 😃

web/models/user.ex
```diff
@@ -19,7 +19,7 @@ defmodule Pairmotron.User do
   @minimum_password_length 8
 
   @required_params ~w(name email)
-  @optional_params ~w(password password_confirmation active role_id)
+  @optional_params ~w(password password_confirmation active is_admin)
```
mbramson: So this made me just think about something fun. I think that any user can create an admin user by manually hitting the create route... Not new to this PR, this just made me think about it. Perhaps we should add a protection around this in the user :create action validating that the logged in user is an admin?

web/models/role.ex
```diff
@@ -1,22 +0,0 @@
...
-  @doc """
-  Builds a changeset based on the `struct` and `params`.
-  """
-  def changeset(struct, params \\ %{}) do
-    struct
-    |> cast(params, [:name, :is_admin])
-    |> validate_required([:name, :is_admin])
-    |> unique_constraint(:name)
-  end
-end
```
mbramson, line 22: 😆

mbramson (approved):
This is awesome. LGTM. One comment on a vulnerability that technically isn't in the scope of this PR as it existed before, but should probably be resolved asap.

web/models/user.ex
ericworkman: I think by having registration_changeset, we are protected from that.

mbramson: Totally correct (obviously just discussed all of this), but the :update action uses the regular changeset, in which you can change the role_id. So a regular user can essentially update themselves so that they are an admin.

ericworkman: Totally an issue with :update

web/models/user.ex
```diff
@@ -19,7 +19,7 @@ defmodule Pairmotron.User do
   @minimum_password_length 8
 
   @required_params ~w(name email)
-  @optional_params ~w(password password_confirmation active is_admin)
+  @optional_params ~w(password password_confirmation active)
```
mbramson, line 22: Perfect!

mbramson (approved):
LGTM now that users can't just tell pairmotron that they are now admins

PR #66, opened 2017-01-26
Refactor GroupController and PairRetroController to use @authorized_conn

mbramson:
Let me know what you think of this refactor. If this style makes sense and makes it more readable, I'd like to apply it to other controllers. Thoughts?


ericworkman:
I like this!

mbramson:
@daveshah showed me that it is in fact possible to import @authorized_conn from a different module, so I'm going to get that working before this is merged.

mbramson:
Managed to extract @authorized_conn to web/web.ex with little issue, and also apply this refactor to the PairRetroControler as well. This should be good to go.

ericworkman (approved):
LGTM I definitely like this way better than the previous

PR #68, opened 2017-01-27
Add User Profile

### LINKED ISSUE #67: Create a /profile route which shows the current user's information
This would be useful so that we can remove most (or all) of the existing /users routes and rely only on ex_admin for administrative user modification.

mbramson:
Add User Profile. This allows users to manage their own settings and see which groups they are a part of.

Removes User routes, controller, views, and templates, as all of this functionality is capture now by the user profile and ex_admin.

TODO:

- [x] Add the groups that a user is associated with
- [x] In the same displays as the groups, display when the user is owner of a group

Closes #67


test/controllers/user_controller_test.exs
```diff
@@ -81,7 +81,6 @@ defmodule Pairmotron.UserControllerTest do
 
     test "does not update chosen resource and renders errors when data is invalid", 
       %{conn: conn, logged_in_user: user} do
-      conn = log_in(conn, user)
```
mbramson: This was already performed in the setup, looks like a previous oversight.

test/support/factory.ex
```diff
@@ -44,8 +44,9 @@ defmodule Pairmotron.Factory do
 
   def group_factory do
     %Group{
-      name: sequence(:name, &"name #{&1}"),
-      owner: build(:user)
+      name: sequence(:name, &"group #{&1}"),
+      owner: build(:user),
+      users: []
```
mbramson, line 49: Although nothing was failing because this wasn't specified, this should make the factory a bit more robustness in the future.

web/controllers/registration_controller.ex
```diff
@@ -6,11 +6,11 @@ defmodule Pairmotron.RegistrationController do
   plug :scrub_params, "user" when action in [:create]
 
   def new(conn, _params) do
-    render conn, changeset: User.registration_changeset(%User{})
+    render conn, changeset: User.profile_changeset(%User{})
   end
 
   def create(conn, %{"user" => user_params}) do
-    changeset = User.registration_changeset(%User{}, user_params)
+    changeset = User.profile_changeset(%User{}, user_params)
```
mbramson, line 13: So I was creating a third user changeset called `profile_changeset`, and I realized that it looked exactly like `registration_changeset`. Rather than just reuse registration_changeset when there is no registration taking place, I thought it made more sense to just use profile_changeset as the name for both.

web/router.ex
```diff
@@ -46,6 +46,9 @@ defmodule Pairmotron.Router do
     pipe_through [:browser, :authenticate] # Use the default browser stack
 
     get "/pairs", PageController, :index
+    get "/profile", ProfileController, :show
+    get "/profile/edit", ProfileController, :edit
+    put "/profile/update/:id", ProfileController, :update
```
mbramson: So I'd love some advice on this, I couldn't get this route to work without specifying `/:id` in the route. I would've liked to just exclude it, as it always updated the currently logged in user. the `id` parameter is completely ignored and is unnecessary.

test/controllers/pair_retro_controller_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule Pairmotron.PairRetroControllerTest do
   @invalid_attrs %{}
 
   test "redirects to sign-in when not logged in", %{conn: conn} do
-    conn = get conn, user_path(conn, :index)
+    conn = get conn, pair_retro_path(conn, :index)
```
mbramson, line 12: Oops...

ericworkman (approved):
LGTM awesome job!

PR #75, opened 2017-01-29
Add Groups and UserGroups to the admin interface

ericworkman:
@mbramson Add Groups and UserGroups to the admin interface with appropriate actions and filters for each. This also lets you play with adding users to others' groups.


mbramson (approved):
Awesome. LGTM.

PR #76, opened 2017-01-29
Redesign navbar to be an actual navbar

### LINKED ISSUE #48: New navbar design
cc: @mbramson 

- Change appearance of login/logout buttons
- Move navbar to its own line

ericworkman:
Make the navbar a real static navbar
Include the collapse.js functionality (and more) from bootstrap
Fix some css mostly around logo and different margins

New design:
screen shot 2017-01-29 at 3 10 04 pm

Mobile browsers now work with a collapse:
screen shot 2017-01-29 at 3 10 24 pm

Closes #48 

@mbramson Eventually I think the Logout link and the Profile link can go into a dropdown (with the dropdown text being the user's name), but we have a bunch of tests that test for inclusion of a user's name on the page. We'll have to improve those tests and probably decide that more than 2 things go into the dropdown before we make that change.

### DIFF
--- a/brunch-config.js
+++ b/brunch-config.js
@@ -8,20 +8,6 @@ exports.config = {
         "js/admin_lte2.js": ["web/static/vendor/admin_lte2.js"],
         "js/jquery.min.js": ["web/static/vendor/jquery.min.js"]
       }
-      // To use a separate vendor.js bundle, specify two files path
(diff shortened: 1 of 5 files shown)


ericworkman:
Also this design is wider than the previous design.

mbramson (approved):
Excellent! LGTM

PR #78, opened 2017-01-30
Adjust pairs route

mbramson:
Currently the pairs route for a specific week is /year/week. This is a bit unintuitive since currently, the current week's pairs are at /pairs. I adjusted the specific week route so that it was /pairs/year/week.

I also fixed a number of elixir 1.4 warnings around functions without () that are ambiguous.


ericworkman (approved):
LGTM! Thanks for pointing out the inconsistency of the /year/week route and fixing it!

PR #79, opened 2017-01-30
Add :show action to allowable actions for user_groups admin

ericworkman:
The result of the :new action is a redirect to the :show action, which shows a 403 route if :show is not allowed.

@mbramson This is the fix for the bug you found.


mbramson (approved):
LGTM

PR #80, opened 2017-02-01
73: Make projects group specific

mbramson:
Adds a group relation to projects so that users in groups can have projects that are private to that group.

Users can create projects as long as they are associated with at least one group. If they are not they are directed to a screen where they can either create or find a different group (I realized actually joining someone else's group is not implemented yet).

Once a project has been created, the group that it's associated with can never be changed (except with /admin).

TODO:

- [x] Setup Project form to allow a selection of the Group to be associated with a project.
  - [x] This dropdown should only list groups that the current user is associated with.
- [x] Make selection of group only available when creating a project
  - [x] for the update route
  - [x] for the edit form
- [x] Disable the ability to hit the project :create action with a group_id that the current user is not associated with. (i.e. add validation to the changeset)
- [x] Make the group required upon creation and update
- [x] If a user isn't a member of any groups tell them that they have to join/create a group in order to make a project since projects have to be associated with groups.

### DIFF
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -9,26 +9,25 @@ defimpl Canada.Can, for: Pairmotron.User do
   def can?(%User{id: user_id}, action, %Group{owner_id: user_id})
     when action in [:edit, :update, :delete], do: true
 
-  def can?(%User{id: user_id}, :show, project = %Project{}) do
-    if is_nil(project.group_id) do
-      project = project |> Pairmotron.Repo.preload([{:pair_retros, :user}])
-      project.pair_retros
-      |> Enum.any?(fn retro -> retro.user.id == user_id end)
-    else
-      project = project |> Pairmotron.Repo.preload([{:group, :users}])
-      project.group.users
-      |> Enum.any?(fn user -> user.id == user_id end)
-    end
+  def can?(%User{id: user_id}, :show, project = %Project{group_id: nil}) do
+    project = project |> Pairmotron.Repo.preload([{:pair_retros, :user}])
+    project.pair_retros
+    |> Enum.any?(fn retro -> retro.user.id == user_id end)
   end
+
+  def can?(%User{id: user_id}, :show, project = %Project{}) do
+    project = project |> Pairmotron.Repo.preload([{:group, :users}])
+    project.group.users
+    |> Enum.any?(fn user -> user.id == user_id end)
+  end
+
+  def can?(%User{}, action, %Project{group_id: nil})
+    when action in [:edit, :update, :delete], do: false
 
   def can?(%User{id: user_id}, action, project = %Project{})
     when action in [:edit, :update, :delete] do
-    if is_nil(project.group_id) do
-      false
-    else
-      project = project |> Pairmotron.Repo.preload(:group)
-      project.group.owner_id == user_id
-    end
+    project = project |> Pairmotron.Repo.preload(:group)
+    project.group.owner_id == user_id
   end
 
   def can?(%User{}, _, _), do: false
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -11,7 +11,9 @@ defimpl Canada.Can, for: Pairmotron.User do
 
   def can?(%User{id: user_id}, :show, project = %Project{}) do
     if is_nil(project.group_id) do
-      false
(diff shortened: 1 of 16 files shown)


mbramson:
@ericworkman take a look when you get the chance. This is feature complete now.

lib/abilities.ex
```diff
@@ -9,5 +9,15 @@ defimpl Canada.Can, for: Pairmotron.User do
...
 
+  def can?(%User{id: user_id}, action, project = %Project{})
+    when action in [:edit, :update, :delete] do
+    if is_nil(project.group_id) do
+      false
+    else
+      project = project |> Pairmotron.Repo.preload(:group)
+      project.group.owner_id == user_id
+    end
+  end
```
mbramson: Only the owner of a group can edit a project associated with that group. Eventually this will be expanded to include users with the relevant GroupRoles, but for now owners are the only ones authorized.

priv/repo/migrations/20170201015913_add_group_to_project.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Pairmotron.Repo.Migrations.AddGroupToProject do
+  use Ecto.Migration
+
+  def change do
+
+    alter table(:projects) do
+      add :group_id, references(:groups, on_delete: :nilify_all)
+    end
```
mbramson, line 8: As discussed, deleting a group does not delete associated projects (although they will end up in an orphaned state). This is so that users don't lose data on things that they have paired on in the past.

lib/abilities.ex
```diff
@@ -9,5 +9,25 @@ defimpl Canada.Can, for: Pairmotron.User do
...
 
+  def can?(%User{id: user_id}, :show, project = %Project{}) do
+    if is_nil(project.group_id) do
+      false
+    else
+      project = project |> Pairmotron.Repo.preload([{:group, :users}])
+      project.group.users
+      |> Enum.any?(fn user -> user.id == user_id end)
+    end
+  end
```
mbramson: So the thought here is that if a project is disassociated from a group because the group is deleted, users still can't look at it. Does this make sense?

web/models/project.ex
```diff
@@ -23,6 +24,23 @@ defmodule Pairmotron.Project do
     |> validate_url(:url)
   end
```
mbramson, line 25: So the default changeset exists basically only to let the /admin route change whatever they want. User facing routes use only `changeset_for_create` and `changeset_for_update`

mbramson (commented):
Some thoughts on the PR

lib/abilities.ex
mbramson: So I added the edge case here, where users are allowed to look at a project that does not have a group so long as they have a retrospective that is associated with that project.

There is also an edge case to that, where they shouldn't be able to see a project that they have a retrospective associated with if that project is associated with a group that the user is no longer in. This is to prevent a user that has been removed from a group from seeing anything they aren't supposed to, as a project's resources should be private to members of that group.

mbramson (commented):
Comments on an edge case.

lib/abilities.ex
ericworkman: 👍

lib/abilities.ex
ericworkman: Is it possible to split this into two functions?

`def can?(%User{id: user_id}, :show, project = %Project{group_id: group_id}) when is_nil(group_id) do`?

ericworkman: If a user were to ever edit a retro that includes a project that no longer belongs to a group of the user, will the user be able to save the changes? Would we want the user to be able to save the changes?

Oh, this is handled by not allowing the user to edit the group, right?

ericworkman (commented):
Looking good!

lib/abilities.ex
mbramson: Yep, totally possible. I didn't do it because you can't access struct properties in guard clauses, but forgot that I can just pattern match on the property directly. Not the first time I've forgotten that.

So since a retro is tied to a specific user, I think that it's probably okay to let someone edit all of their own retros, pretty much always. That's definitely something to think about though.

mbramson:
Refactored lib/abilities to be more idiomatic, and betterer in general. This is done at this point.

ericworkman (approved):
LGTM!

PR #81, opened 2017-02-03
Refactor repairify

ericworkman:
@mbramson I wanted to get this down for you to at least look at and to track remaining work. I tried using structs to somewhat structure the data that the builder and pairer return. This more or less works, but it has some odd edge cases and missing features.

- [x] Ignore pairs that have retro'd when finding available pairs for single users
- [x] Decide the boundaries and responsibilities of these new things
- [x] Flesh out some of the more complicated test cases, especially for the Pairer
- [x] Settle on the proper structs for PairBuilder and Pairer
        Happy enough with these.
- [x] Account for the case where the sole existing pair already has 3 members


ericworkman:
@mbramson Leaving off the pipeline stuff we discussed. I'm calling this complete for now. I may go back and refactor and rename some things in the future. As long as neither of us has better names, these are staying. 

I made a couple decision in regard to pair length. I decided to never increase a 3-pair, not to break up 3-pairs just to pair with a new user, and that 1-pairs are okay if there are no other options. If the 1-pair (with an existing 3-pair) isn't okay, an admin can go in and inactive a user in the 3-pair, repairify, reactivate the user, and then repairify.

I encourage to to try out as many weird test cases as you can.

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  @doc """
+  Finds the previous %UserPair{}s that will no longer be needed with the new user set.
+  """
+  def find_dead_user_pairs([], _), do: []
+  def find_dead_user_pairs(user_pairs, []), do: user_pairs
+  def find_dead_user_pairs(user_pairs, users) do
+    previously_paired_users = user_pairs |> pair_users |> MapSet.new
+    u = users |> MapSet.new
```
mbramson: Instead of `previously_paired_users` and `u`, how about `previously_paired_users_set` and `current_users_set`?

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+    dead_users = MapSet.difference(previously_paired_users, u) |> MapSet.to_list
+    dead_user_pairs = dead_users
+      |> Enum.reduce([], fn(dead_user, dup) ->
+        dup ++ Enum.filter(user_pairs, fn(p) -> p.user == dead_user end)
+      end)
```
mbramson: Hmmm... This seems wasteful. If I'm not mistaken this is O(n^2), and also made a bit slower than necessary because ++ traverses the accumulator every time it's used. This also took me a lot longer than it should have to grok. How about:

```
dead_user_pairs = dead_users
|> Enum.map(fn dead_user -> Enum.find(user_pairs, &(&1.user.id == dead_user.id)) end)
|> Enum.filter(&(!is_nil(&1)))
```

This still isn't quite ideal, there may be some further optimization here if we make some model association tradeoffs, but it's definitely more performant. Since we're doing a find rather than a filter within the reduce, we don't traverse the entire list (although we still do worse case), we short circuit when we first find a match. This also gets rid of the ++ operator. I think this reads better too.

The filter that I wrote might seem like it's unnecessary, but it serves two subtly important purposes. It lessens the degree to which we have to trust input, and it guards against a race condition since users and user_pairs are retrieved from the database in different transactions.

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+    dead_matched_user_pairs = dead_user_pairs
+      |> Enum.reduce([], fn(dead_user_pair, dead_matched_user_pairs) ->
+        dead_matched_user_pairs ++ Enum.filter(user_pairs, fn(p) -> p.pair == dead_user_pair.pair && p.id != dead_user_pair.id end)
+      end)
```
mbramson: Similarly, how about:

```
dead_matched_user_pairs = dead_user_pairs
  |> Enum.map(fn dead_user_pair -> Enum.find(user_pairs, &(&1.pair.id == dead_user_pair.pair.id && &1.id != dead_user_pair.id)) end)
  |> Enum.filter(&(!is_nil(&1)))
```

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+    dead_user_pairs ++ dead_matched_user_pairs |> Enum.uniq
```
mbramson: Is the Enum.uniq necessary here? I think that the second half of the boolean statement in the second Enum.find (`p.id != dead_user_pair.id`) basically ensures that there would be no duplicates here.

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  defp find_users_to_pair(dead_user_pairs, user_pairs, users) do
+    unchanged_users = find_remaining_user_pairs(dead_user_pairs, user_pairs)
+      |> pair_users
+      |> MapSet.new
+    users
+      |> MapSet.new
+      |> MapSet.difference(unchanged_users)
+      |> MapSet.to_list
+  end
```
mbramson: I would move this method up to line 61, so that it can be with the other methods that use it.

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  @doc """
+  Finds the %Users{}s that will need to be paired because either their previous pair is no longer
+  valid or the were not previously paired.
+  """
+  def find_users_to_pair(_, []), do: []
+  def find_users_to_pair([], users), do: users
+  def find_users_to_pair(user_pairs, users) do
+    find_users_to_pair(find_dead_user_pairs(user_pairs, users), user_pairs, users)
+  end
```
mbramson: Why are these methods here? I can't seem to see where they're referenced outside of the tests.

web/controllers/pair_controller.ex
```diff
@@ -63,14 +57,45 @@ defmodule Pairmotron.PairController do
...
       |> order_by(:id)
       |> Repo.all
+
+    pair_ids = fetch_pairs(year, week)
+      |> Enum.map(fn(p) -> p.id end)
+
+    query = from up in UserPair,
+            select: up,
+            where: up.pair_id in ^pair_ids,
+            preload: [:pair, :user]
```
mbramson: This query might belong on the UserPair model itself.

web/controllers/pair_controller.ex
```diff
@@ -63,14 +57,45 @@ defmodule Pairmotron.PairController do
...
       |> Mixer.mixify(week)
-      |> Pairer.generate_pairs
+      |> Pairer.generate_pairs(pairs)
+
+    results.pairs
       |> Enum.map(fn(users) -> make_pairs(users, year, week) end)
       |> List.flatten
       |> Enum.map(fn(p) -> Repo.insert! p end)
+
+    insert_user_pair(results.user_pair)
```
mbramson, line 85: I think that this is the perfect place to use `Ecto.Multi` to bring the delete and the inserts into one database transaction to stop us from partially applying changes.

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,94 @@
...
+  and the new list of %User{}s.
+  """
+  def determify(user_pairs, users) do
+    dead_user_pairs = find_dead_user_pairs(user_pairs, users)
+    %Determination{
+      dead_pairs: dead_user_pairs |> find_dead_pairs,
+      remaining_pairs: dead_user_pairs |> find_remaining_user_pairs(user_pairs) |> unique_pairs,
+      available_users: dead_user_pairs |> find_users_to_pair(user_pairs, users)
+    }
+  end
```
mbramson: Why was the decision made to make the arguments `user_pairs` and `users` rather than `pairs` and `users`? Correct me if I'm wrong, but is there ever a case where you want to only delete some of the `UserPair`s associated with a pair? If not, couldn't we just delete the pairs that are no longer valid and rely on the database constraints to take care of the `UserPair` records that need to be deleted?

I feel like this would remove a lot of the logic you had to use to relate `Pair`, `UserPair` and `User` records in various ways, as you can reference the `:users` association on the pairs directly to free up users. I believe that this could be a nice performance boost too, but it's hard to say without doing it. I also think this would remove a few methods entirely and generally be easier to understand.

I wouldn't say change it for this PR, because this is working well, but that's definitely a candidate for a possible refactor down the line.

mbramson (commented):
Some thoughts on various things. I can tell this was a hairy PR, so good work!

ericworkman:
@mbramson Excellent call outs 👍  I've refactored the builder in line with your comments and a refreshed look.

web/controllers/pair_controller.ex
ericworkman: Good thought. I will look into Multi.

lib/pairmotron/pair_builder.ex
```diff
@@ -4,91 +4,62 @@ end
...
+      |> pair_users
+      |> MapSet.new
+    dead_users = previously_paired_users_set
+      |> MapSet.difference(users_set)
+      |> MapSet.to_list
+    dead_users
+      |> Enum.map(fn dead_user -> Enum.find(pairs, &(dead_user in &1.users)) end)
+      |> Enum.filter(&(!is_nil(&1)))
+      |> Enum.uniq
   end
```
mbramson: This is miles cleaner and way more straightforward to understand. Awesome!

test/lib/pair_builder_test.exs
```diff
@@ -1,123 +1,68 @@
 defmodule Pairmotron.PairBuilderTest do
   use ExUnit.Case, async: true
 
-  alias Pairmotron.{PairBuilder, User, Pair, UserPair}
```
mbramson: Pumped we don't even have to think about `UserPair`s in these tests anymore

web/controllers/pair_controller.ex
```diff
@@ -61,18 +61,10 @@ defmodule Pairmotron.PairController do
...
 
-    pair_ids = fetch_pairs(year, week)
-      |> Enum.map(fn(p) -> p.id end)
+    pairs = fetch_pairs(year, week)
+      |> Repo.preload(:users)
 
-    query = from up in UserPair,
-            select: up,
-            where: up.pair_id in ^pair_ids,
-            preload: [:pair, :user]
```
mbramson: Yes! This disappeared entirely!

mbramson (approved):
Fantastic work! That refactor really simplified things. LGTM, let's get this merged.

web/controllers/pair_controller.ex
```diff
@@ -63,14 +57,37 @@ defmodule Pairmotron.PairController do
...
+      |> Repo.preload(:users)
+
+    determination = PairBuilder.determify(pairs, users)
+
+    determination.dead_pairs
+      |> Enum.map(fn(p) -> Repo.delete! p end)
+
+    pairs = determination.remaining_pairs
+      |> Repo.preload([:users, :pair_retros])
+      |> Enum.filter(fn(p) -> length(p.pair_retros) == 0 end)
```
mbramson: Tiny optimization here, instead of length == 0 can we do Enum.any? which will short circuit on the first element rather than iterate through all pair_retros. I realize this is a micro-optimization, but I think it might read better too.

`length(p.pair_retros) == 0` vs `Enum.any?(p.pair_retros)`

mbramson (commented):
Another thought

web/controllers/pair_controller.ex
ericworkman: Oooh

lib/pairmotron/pair_builder.ex
```diff
@@ -0,0 +1,65 @@
+defmodule Determination do
+  defstruct dead_pairs: [], remaining_pairs: [], available_users: []
```
mbramson: Can we give this a module_doc that spells out exactly what each field is, and what it's intended to be used for?

web/controllers/pair_controller.ex
```diff
@@ -63,14 +57,37 @@ defmodule Pairmotron.PairController do
...
       |> order_by(:id)
       |> Repo.all
+
+    pairs = fetch_pairs(year, week)
+      |> Repo.preload(:users)
+
+    determination = PairBuilder.determify(pairs, users)
+
+    determination.dead_pairs
+      |> Enum.map(fn(p) -> Repo.delete! p end)
```
mbramson, line 70: Could this use `Repo.delete_all` instead of of `Repo.delete!` on each pair? `delete_all` would be one trip to the db for all pairs whereas Repo.delete! is a trip to the db for each pair.

ericworkman: I'll tackle this another time. Thanks for the reminder of `delete_all`

PR #87, opened 2017-02-03
Add created_by to project model and db table

### LINKED ISSUE #86: Add created_by association to projects
Add a created_by association on projects that specify who created the project originally.

Right now once a project is created, the only person that can edit a project is the owner of the group that the project is associated with. This is partially good, because we want to protect someone's project, but this causes issues in the situation where a user creates a project, but makes a mistake (or later wants to change the description or something).

The authorization for editing/updating/deleting a project should be changes to additionally allow the creator of the project to do these things.

On the projects index, the user that created a given project should also be able to see the edit/delete buttons which are currently only seen by the owner of the group associated with a given project.

mbramson:
Adds a created_by field to the project model so that the original creator of a project can edit it in the future in addition to the owner of the project's associated group.

TODO:
- [x] Setup Model/DB Migration
- [x] Set created_by upon creation of a project
- [x] Allow the creator of a project to edit/update/delete if they are still in the group
- [x] Ensure that the creator of a project cannot edit/update/delete if they are no longer in the group
- [x] Enable the edit and delete buttons on the project index if the current user is the creator of that specific project even if they are not the owner of the associated group.

Closes #86

### DIFF
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -26,8 +26,12 @@ defimpl Canada.Can, for: Pairmotron.User do
 
   def can?(%User{id: user_id}, action, project = %Project{})
     when action in [:edit, :update, :delete] do
-    project = project |> Pairmotron.Repo.preload(:group)
-    project.group.owner_id == user_id
+    project = project |> Pairmotron.Repo.preload([{:group, :users}])
+    cond do
+      project.group.owner_id == user_id -> true
+      project.created_by_id == user_id -> true
+      true -> false
+    end
   end
 
   def can?(%User{}, _, _), do: false
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -29,6 +29,7 @@ defimpl Canada.Can, for: Pairmotron.User do
     project = project |> Pairmotron.Repo.preload([{:group, :users}])
     cond do
       project.group.owner_id == user_id -> true
+      !Enum.any?(project.group.users, &(&1.id == user_id)) -> false
       project.created_by_id == user_id -> true
       true -> false
     end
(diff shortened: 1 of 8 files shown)


mbramson:
@ericworkman take a look when you get the chance, this is complete!

lib/abilities.ex
```diff
@@ -26,8 +26,13 @@ defimpl Canada.Can, for: Pairmotron.User do
 
   def can?(%User{id: user_id}, action, project = %Project{})
     when action in [:edit, :update, :delete] do
-    project = project |> Pairmotron.Repo.preload(:group)
-    project.group.owner_id == user_id
+    project = project |> Pairmotron.Repo.preload([{:group, :users}])
+    cond do
+      project.group.owner_id == user_id -> true
+      !Enum.any?(project.group.users, &(&1.id == user_id)) -> false
```
ericworkman, line 32: Instead of `!Enum.any?` would `Enum.empty?` work?

ericworkman (commented):
Any plans on changing existing projects to include an owner?

lib/abilities.ex
mbramson: great call!

mbramson: Looks like Enum.empty? doesn't accept a function to verify with, so unfortunately it looks like I have to use !Enum.any?

mbramson:
Also @ericworkman just noticed you asked a question in your last review. I wasn't going to, as it's not necessary per se. I can always go in through /admin and add the proper creators to the various projects (most of them are me at this point).

ericworkman (approved):
LGTM!

Issue #83, opened 2017-02-03
Implement group-specific pairs


mbramson:
Pairs should not be global, they should be filtered by group. This might mean that the route to get to pairs changes.

To get to the pairs for a specific group, the route should be /groups/:group_id/pairs. I don't think that this should deprecate the /pairs route, perhaps this just redirects to the first group's (or some default group specified by the user in the future) pairs.

This will involve adding an association between pairs and groups.

This work should be decoupled from the work to redesign the pairing engine currently being spearheaded by @ericworkman.

ericworkman:
Completed by #88

PR #88, opened 2017-02-04
Associate pairs with groups

ericworkman:
@mbramson 
Depends on and merges into #81 . Pairs now depend on groups. I had to change a lot of redirects and tests for this. I created a conflict with "create_pair/2" that was resolved by using the full namespace. Most back links now redirect to the profile page. I added a link to the table of groups on the profile page for the pairs for each group.

I'll probably work on beatifying some of these things soon sans react.

- [x] New /pairs
- [x] Authz for /group/:id/pairs...


test/support/test_helper.ex
```diff
@@ -55,9 +55,16 @@ defmodule Pairmotron.TestHelper do
...
     retro = create_retro(user, pair)
     {pair, retro}
   end
+
+  def create_group(user) do
+    group = Group.changeset(%Group{}, %{owner_id: user.id, name: "#{DateTime.utc_now()}"})
+      |> Repo.insert!
+    UserGroup.changeset(%UserGroup{}, %{user_id: user.id, group_id: group.id}) |> Repo.insert!
+    group
+  end
```
mbramson: Everything here can be replaced with ex_machina in the tests.

test/controllers/pair_controller_test.exs
```diff
@@ -1,110 +1,108 @@
...
-      conn = get(conn, "/pairs")
+      conn = get(conn, "/groups/#{group.id}/pairs")
       refute html_response(conn, 200) =~ user.name
     end
 
-    test "pairs two users together", %{conn: conn, logged_in_user: user1} do
+    test "pairs two users together", %{conn: conn, logged_in_user: user1, group: group} do
       user2 = insert(:user)
-      conn = get(conn, "/pairs")
+      UserGroup.changeset(%UserGroup{}, %{user_id: user2.id, group_id: group.id}) |> Repo.insert!
```
mbramson: Instead of doing an insert like this, you can just do `user2 = insert(:user, %{group: group})` instead of `user2 = insert(:user)`

test/controllers/pair_controller_test.exs
```diff
@@ -1,110 +1,108 @@
...
       refute Repo.get_by(UserPair, %{user_id: user3.id, pair_id: pair.id})
     end
 
-    test "repairifying does not delete valid pairs", %{conn: conn, logged_in_user: user} do
+    test "repairifying does not delete valid pairs", %{conn: conn, logged_in_user: user, group: group} do
       {year, week} = Timex.iso_week(Timex.today)
       user2 = insert(:user)
-      pair = create_pair([user, user2])
-      delete conn, pair_path(conn, :delete, year, week)
+      UserGroup.changeset(%UserGroup{}, %{user_id: user2.id, group_id: group.id}) |> Repo.insert!
```
mbramson: Same as above, this can be accomplished more elegantly with `user2 = insert(:user, %{group: group})`

web/controllers/pair_controller.ex
```diff
@@ -1,52 +1,55 @@
...
   use Pairmotron.Web, :controller
 
-  alias Pairmotron.{Pair, PairRetro, User, UserPair, Mixer, Pairer, PairBuilder}
+  alias Pairmotron.{Group, Pair, PairRetro, UserPair, Mixer, Pairer, PairBuilder}
 
-  def index(conn, _params) do
+  def index(conn, %{"group_id" => g}) do
     {year, week} = Timex.iso_week(Timex.today)
-    pairs = fetch_or_gen(year, week)
+    {gid, _} = g |> Integer.parse
```
mbramson: Why not just spell out `group_id`?

web/router.ex
```diff
@@ -45,18 +45,18 @@ defmodule Pairmotron.Router do
   scope "/", Pairmotron do
     pipe_through [:browser, :authenticate] # Use the default browser stack
 
-    get "/pairs", PairController, :index
```
mbramson: So see https://github.com/ElixirCLE/pairmotron/issues/83, but I don't think that we should actually get rid of the `/pairs` route, I just think that it should be repurposed. I think that all of these redirects that are now redirecting to the profile are the wrong course of action. I still think that the user should see pairs in these cases, not their profile as pairs is really the central function of the app.

So I'd make the `/pairs` index just redirect to `/groups/:group_id/pairs` where the group_id at this point is just the group that the user is a member of with the lowest id (for now). In the future it'd be nice to let the user set which group they'd like to default to.

This also lowers the amount of changes you have to make in this PR (although I realize you've already made them).

mbramson (changes requested):
Some thoughts on this PR. What do you think about the /pairs route?

web/router.ex
ericworkman: I don't like having `/pairs` redirect to one of the user's group pair pages. I think its surprising in the best case. It would be less surprising and more useful to have `/pairs` include the pairs for each of the user's groups for that time period. I'm not going to implement that unless we both think its a good idea.

ericworkman:
@mbramson Made a lot of the changes you requested. Still open on the usage of `/pairs`.

ericworkman:
Discussed this with @mbramson. As part of this PR, I will be making a new "/pairs" that will display the current user's pair within each of their groups. Then most of the redirect changes will change back to redirects to "/pairs". I will also be adding an authorization check to the group pairs route to verify that a user is in the group they are trying to view.

ericworkman:
Still doing:

- [x] Authorization check for GroupPairController
- [x] Switch most of the redirects back to /pairs
- [x] Refactor some of the shared things between GroupPairController and PairController
- [x] Refactor PairController to remove duplicate logic between show and index
- [x] Move PairMaker specific tests out of the controller tests
- [x] Double check the need for any migrations for existing pairs
- [x] Fix issue with /pairs showing more than its supposed to

ericworkman:
@mbramson Review please. 
Here's what the current /pairs looks like for a user:
screen shot 2017-02-16 at 5 04 30 pm
Versus the "The Oh Yeahs" group's pairs for the same period:
screen shot 2017-02-16 at 5 04 46 pm

The added migration will associate all current pairs to the only existing group in prod. For dev environments it could be a little weird if the first group isn't the right group for users. Worst case is that the dev db needs to be cleaned up. Any suggestions for improving that migration so that it is smarter?

Should we add the group's name to the group pair page?

web/controllers/group_pair_controller.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Pairmotron.GroupPairController do
+  use Pairmotron.Web, :controller
+
+  import Pairmotron.ControllerHelpers
+  alias Pairmotron.{PairMaker, Group}
+
+  plug :load_and_authorize_resource, model: Group, only: [:index, :show]
```
mbramson: Why :index in addition to :show? There's no can? protocol defined and no index action on this controller.

web/controllers/pair_controller.ex
```diff
@@ -2,108 +2,36 @@ defmodule Pairmotron.PairController do
...
+    user = user
+      |> Repo.preload(:groups)
+    pairs = user.groups
+      |> Enum.map(fn(g) ->
+        PairMaker.fetch_or_gen(year, week, g.id)
+          |> Enum.filter(fn(p) ->
+               member_ids = p.users |> Enum.map(&(&1.id))
+               Enum.member?(member_ids, user.id)
+             end)
+      end)
```
mbramson: Can we be explicit with our temp variables? So `group` instead of `g` and `pair` instead of `p`?

web/controllers/pair_controller.ex
```diff
@@ -2,108 +2,36 @@ defmodule Pairmotron.PairController do
...
+  defp fetch_pairs_per_group(year, week, user) do
+    user = user
+      |> Repo.preload(:groups)
+    pairs = user.groups
+      |> Enum.map(fn(g) ->
+        PairMaker.fetch_or_gen(year, week, g.id)
+          |> Enum.filter(fn(p) ->
+               member_ids = p.users |> Enum.map(&(&1.id))
+               Enum.member?(member_ids, user.id)
+             end)
```
mbramson: Could we extract lines 29 through 33 to their own method called something like `users_pair_for_group(user, year, week, group)`. This code is about as elegant as it can be without extraction, but I think extraction would improve the ease of understanding both `fetch_pairs_per_group` and the current in-line block to be extracted.

lib/abilities.ex
```diff
@@ -9,6 +9,12 @@ defimpl Canada.Can, for: Pairmotron.User do
   def can?(%User{id: user_id}, action, %Group{owner_id: user_id})
     when action in [:edit, :update, :delete], do: true
 
+  def can?(user = %User{}, :show, group = %Group{}) do
+    group = group |> Pairmotron.Repo.preload(:users)
+    group.users
+      |> Enum.any?(fn guser -> guser.id == user.id end)
```
mbramson: Could lines 14-15 be simplified to `Enum.member?(group.users, user)`

web/router.ex
```diff
@@ -43,20 +43,23 @@ defmodule Pairmotron.Router do
...
     get "/profile/edit", ProfileController, :edit
     put "/profile/update/:id", ProfileController, :update
     resources "/projects", ProjectController
+
+    get "/pairs", PairController, :index
+    get "/pairs/:year/:week", PairController, :show
+
+    get "/groups/:id/pairs", GroupPairController, :show
+    get "/groups/:id/pairs/:year/:week", GroupPairController, :show
+    delete "/groups/:group_id/pairs/:year/:week", GroupPairController, :delete
```
mbramson: The name of the group_id variable is different for these three routes. I think `:group_id` makes the most sense, but they should probably be consistent.

web/templates/group_pair/index.html.eex
```diff
@@ -0,0 +1,20 @@
+<h2>Pairs for week <%= @week %> of <%= @year %></h2>
+<div class="row">
+  <div class="col-md-8 col-sm-12 col-xs-12">
+    <ul class="list-group">
+      <%= for pair <- @pairs do %>
+        <%= render Pairmotron.SharedView, "pair.html", pair: pair, conn: @conn %>
```
mbramson: Actually did not realize you could plug and play views like that. Thought it was templates only. Awesome!

lib/abilities.ex
mbramson: Also, I'm kind of worried about this being attached to the `%Group{}` can? when it is used on `GroupPairController`, as the `GroupController` uses the `%Group{}` can? implementation. I realize that they use different actions and so this fits nicely in here for now, but what if I decide that a Group :show action should be visible to someone different and want to expand the can? implementation?

Practical example of such an expansion includes maybe wanting users that have received invites for groups to be able to view the :show route in addition to only users in those groups (say if the group were private). You wouldn't want to also give invited users the option see the pairs for that group (maybe? potentially? just an example). In that case there would be collision for the two :show can? implementations, and I would think the one actually going to the Group itself's show action should win out.

Unfortunately, this is a limitation of Canary/Canada. I think that this permission check should be in the GroupPairController show action itself :(.

test/controllers/group_pair_controller_test.exs
```diff
@@ -0,0 +1,101 @@
...
+      group = insert(:group, %{owner: user, users: [user]})
+      conn = build_conn() |> log_in(user)
+      {:ok, [conn: conn, logged_in_user: user, group: group]}
+    end
+
+    test "lists one active user", %{conn: conn, logged_in_user: user, group: group} do
+      conn = get(conn, "/groups/#{group.id}/pairs")
+      assert html_response(conn, 200) =~ user.name
+    end
+
```
mbramson, line 24: What about if there are no users in the group at all? Weirdly enough, this is a possible state right now. You can be the owner of the group and not actually in the users association.

test/controllers/group_pair_controller_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    test "cannot access group pairs", %{conn: conn, group: group} do
+      conn = get conn, group_pair_path(conn, :show, group.id)
+      assert redirected_to(conn) == pair_path(conn, :index)
+    end
+
+    test "cannot access group pairs for a specific period", %{conn: conn, group: group} do
+      conn = get conn, group_pair_path(conn, :show, group.id, 2000, 1)
+      assert redirected_to(conn) == pair_path(conn, :show, 2000, 1)
+    end
```
mbramson, line 116: This might not warrant a test, but should an admin be able to repairify a group that they're not in? I'm not sure what the answer is.

mbramson (changes requested):
Awesome work! I have a few small things that I think should be addressed, but otherwise this looked great!

mbramson:
And also, that's a good point. I think that we should add the group name to the group specific pairs page.

web/controllers/group_pair_controller.ex
ericworkman: Oversight 🙈

web/controllers/pair_controller.ex
ericworkman: Sounds good to me

web/controllers/pair_controller.ex
ericworkman: Sure thing

lib/abilities.ex
ericworkman: > Could lines 14-15 be simplified to `Enum.member?(group.users, user)`

No unfortunately. That was the cause of the last bug I found. The user and the group.users have to match exactly, including preloaded associations and other things, for `Enum.member?` to work.

> I think that this permission check should be in the GroupPairController show action itself :(.

I leaned that way initially too. It fit perfectly as you say, but could be the cause of some trouble later. Want me to remove this now?

web/router.ex
ericworkman: oversight, thanks

test/controllers/group_pair_controller_test.exs
ericworkman: Then you won't even see these pages. Want me to add a test case around this?

test/controllers/group_pair_controller_test.exs
ericworkman: Good call. I'll add a test case to the `describe "while authenticated as admin" do` section.

lib/abilities.ex
mbramson: Ya I think it could also cause confusion down the line. Someone might not know where that rule is actually applied, and be misled about how authorization works for groups.

test/controllers/group_pair_controller_test.exs
mbramson: Up to you. It's definitely an edge case.

ericworkman: I feel like this is handled by the tests of non-group membership access and inactive users.

test/controllers/group_pair_controller_test.exs
ericworkman: Now handled in the "while authenticated as admin" section. 👍

ericworkman:
@mbramson Added the group name to the group_pair page. Ready for more review.

web/router.ex
```diff
@@ -43,20 +43,23 @@ defmodule Pairmotron.Router do
...
     get "/profile/edit", ProfileController, :edit
     put "/profile/update/:id", ProfileController, :update
     resources "/projects", ProjectController
+
+    get "/pairs", PairController, :index
+    get "/pairs/:year/:week", PairController, :show
+
+    get "/groups/:id/pairs", GroupPairController, :show
+    get "/groups/:id/pairs/:year/:week", GroupPairController, :show
+    delete "/groups/:id/pairs/:year/:week", GroupPairController, :delete
```
mbramson, line 58: Why `:id` instead of `:group_id`? As these are routes for `Pair`s, I'd think `:id` would be reserved for a specific pair (if there was ever some route for updating a specific pair for some reason). Therefore, I'd tihnk that `:group_id` would be more appropriate here for the route.

mbramson (approved):
One thought on the name of a variable in some routes, but otherwise LGTM! Fantastic work!

web/router.ex
ericworkman: The `:id` is what allowed canary to match on a `%Group{}`. I agree that `:group_id` is more specific, but I think `:id` matches the general pattern. These are really routes for groups first, then routes for pairs. If we could go the embedded resources way, it would still come through as `:id`.

PR #89, opened 2017-02-04
Add gravatar and make pretty

ericworkman:
@mbramson as it says. Here are some screenshots:

Desktop:
screen shot 2017-02-04 at 5 49 36 pm

IPhone 6 Plus
screen shot 2017-02-04 at 5 49 42 pm

### DIFF
--- a/web/templates/profile/show.html.eex
+++ b/web/templates/profile/show.html.eex
@@ -1,46 +1,48 @@
-<h2>User Profile</h2>
+<div class="row">
+  <div class="col-md-5 col-sm-12">
+    <div class="panel-default panel">
+      <div class="panel-body">
+        <div class="media">
+          <div class="media-left">
+            <img class="img-circle" src="<%= Gravatar.url(@user.email) %>" alt="<%= @user.name %>" height="140" width="140">
+          </div>
+          <div class="media-body">
+            <h2><%= @user.name %></h2>
+            <p><%= @user.email %><p>
+            <p>
+              <%= if @user.active do %>
+                active
+              <%= else %>
+                inactive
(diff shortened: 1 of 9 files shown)


ericworkman:
I've made the /pair route prettier too. I expect to add more to the user panel over time.

screen shot 2017-02-04 at 7 27 31 pm


screen shot 2017-02-04 at 7 27 49 pm
screen shot 2017-02-04 at 7 27 56 pm

web/templates/profile/show.html.eex
```diff
@@ -1,46 +1,48 @@
-<h2>User Profile</h2>
```
mbramson, line 1: Why the decision to drop this off?

mbramson (commented):
I feel like the size of the new icons on the /pairs route sort of deemphasizes the aspects of the page that make the fact that you're paired pop. Does that makes sense? I feel like its harder at first glance to figure out that there's actually pairing going on.

Perhaps smaller icons that are to the left of the user's name would keep the icons from overpowering the rest of the design?

This is awesome! Great idea!

web/templates/profile/show.html.eex
ericworkman: I felt it was redundant. I could keep it as `Profile` if you'd like.

ericworkman:
@mbramson like this better?

screen shot 2017-02-05 at 11 34 38 am
screen shot 2017-02-05 at 11 34 33 am

mbramson:
Yep, that looks solid. LGTM

mbramson (approved):
LGTM

PR #90, opened 2017-02-05
Disable non admin repairify

### LINKED ISSUE #36: Only allow administrators to regenerate pairs for a given week
Currently, anyone can click the "repairify" button to regenerate the pairs for a given week. This should be blocked behind the is_admin parameter of role.

mbramson:
Disable repairifying for users that are not admins

Closes #36

### DIFF
--- a/web/controllers/pair_controller.ex
+++ b/web/controllers/pair_controller.ex
@@ -1,6 +1,7 @@
 defmodule Pairmotron.PairController do
   use Pairmotron.Web, :controller
 
+  import Pairmotron.ControllerHelpers
   alias Pairmotron.{Pair, PairRetro, User, UserPair, Mixer, Pairer, PairBuilder}
 
   def index(conn, _params) do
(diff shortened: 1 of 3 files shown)


web/controllers/pair_controller.ex
```diff
@@ -24,13 +25,17 @@ defmodule Pairmotron.PairController do
...
 
   def delete(conn, %{"year" => y, "week" => w}) do
-    {year, _} = y |> Integer.parse
-    {week, _} = w |> Integer.parse
-    fetch_pairs(year, week)
-      |> Enum.map(fn(p) -> Repo.delete! p end)
-    conn
-      |> put_flash(:info, "Repairified")
-      |> redirect(to: pair_path(conn, :show, year, week))
+    if conn.assigns.current_user.is_admin do
```
mbramson: Since there's no :id parameter in the :index route, canary yells that it's doing something unsafe because it is comparing against nil. Unfortunately, I have to do the manual style admin check here.

web/templates/pair/index.html.eex
```diff
@@ -23,9 +23,11 @@
 </ul>
 <%= if !past_week?(@year, @week) do %>
   <%= form_for @conn, pair_path(@conn, :delete, @year, @week), [as: :repairify, method: "delete"], fn _ -> %>
-    <div class="form-group">
-      <%= submit "Repairify", class: "btn btn-danger" %>
-    </div>
+    <%= if @conn.assigns.current_user.is_admin do %>
```
mbramson, line 27: In the future, I'll probably extract this logic out to the view to account for GroupRoles (which are not yet a thing).

mbramson (commented):
A couple of thoughts

ericworkman (approved):
This looks good to me. I know this will be very useful when we have group admins, but we should consider adding this global admin functionality into the /admin interface.

PR #94, opened 2017-02-06
Fix registration not signing in with Guardian

### LINKED ISSUE #93: Fix redirect issue when signing up for pairmotron
When signing up, when the user submits their information to create a new account and hits submit, the user should be logged in and taken to the /pairs page. There appears to be a bug during that process that results in the following:

register-redirect-bug

mbramson:
Fixes a registration session bug where it was signing in using Plug.conn.put_session rather than Guardian.sign_in.

Closes #93


ericworkman (approved):
LGTM

PR #96, opened 2017-02-17
91 add group membership request

### LINKED ISSUE #91: Implement Request for Membership and Invitations to Groups
Currently, the only way for a user to join a group is through the /admin interface, which is obviously very much less than ideal.

Users should be able to request membership for a group. When they do so, they should be able to see that they have done so somehow (and not be able to request membership again while the first has not been acted upon). The owner of the group that the user has asked to join should be able to see that the user has requested membership (ideally in a centralized location with all requests for membership for that group). They should be able to either deny membership, or accept the request for membership. If they accept, the user should then be associated with that group.

Owners should be able to invite users to be a member of the group that they own. Users should be able to see a list of all the groups that they've been invited to. They should be able to accept or deny the invitation. If they accept, the user should then be associated with that group.

mbramson:
Allow users to request membership in groups and for groups to invite users.

TODO:
- [x] UsersGroupMembershipRequestController
  - [x] :index (user can see all of the outstanding invites they have)
  - [x] :create (user can request membership in a given group)
  - [x] :update (user can accept an invitation that was created by a group owner to join that group)
- [x] GroupInivitationController
  - [x] :index (group owner can see all outstanding invites associated with a group)
  - [x] :new (allow group owner to find a user to invite)
    - [x] Make users dropdown only list users not in the group
  - [x] :create (invites a specific user as the owner of the group)
    - [x] Make errors render to changeset for :new action
  - [x] :update (group owner can accept a membership request that a user created to join their group)
- [x] UI
  - [x] UI for users to request membership in a specific group
  - [x] UI for users to accept membership in a specific group
  - [x] UI for group owners to invite specific users to their group
  - [x] UI for group owners to accept membership requests from a user to their group
  - [x] Link to user's invitations (in profile? dropdown from profile navbar?
  - [x] Link to group's invitations (on group show if current_user is owner of group)

Closes #91 

Subsequent Issues:
- [ ] Allow deletion of group_membership_requests (refusing invitation)

### DIFF
--- a/web/controllers/group_invitation_controller.ex
+++ b/web/controllers/group_invitation_controller.ex
@@ -1,7 +1,7 @@
 defmodule Pairmotron.GroupInvitationController do
   use Pairmotron.Web, :controller
 
-  alias Pairmotron.{Group, GroupMembershipRequest, UserGroup}
+  alias Pairmotron.{Group, GroupMembershipRequest, User, UserGroup}
   import Pairmotron.ControllerHelpers
 
   def index(conn, %{"group_id" => group_id}) do
@@ -16,14 +16,57 @@ defmodule Pairmotron.GroupInvitationController do
     else
       handle_resource_not_found(conn)
     end
-
   end
 
-  def create(conn, %{"group_membership_request" => group_membership_request_params}) do
-    conn
+  def create(conn, %{"group_id" => group_id, "group_membership_request" => group_membership_request_params}) do
+    current_user = conn.assigns.current_user
+
+    {group_id_int, _} = Integer.parse(group_id)
+    group = preloaded_group_or_nil(group_id_int)
+
+    user_id = parameter_as_integer(group_membership_request_params, "user_id")
+    user = Repo.get(User, user_id)
+
+    cond do
+      is_nil(user) ->
+        handle_resource_not_found(conn)
+      is_nil(group) ->
+        handle_resource_not_found(conn)
+      group.owner_id != current_user.id ->
+        redirect_and_flash_error(conn, "You must be the owner of a group to invite user to that group", group_id)
+      user in group.users ->
+        redirect_and_flash_error(conn, "Cannot invite user that is already in group", group_id)
+      true ->
+        implicit_params = %{"initiated_by_user" => false, "group_id" => group_id}
+        final_params = Map.merge(group_membership_request_params, implicit_params)
+        changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{}, final_params)
+
+        case Repo.insert(changeset) do
+          {:ok, _group_membership_request} ->
+            #Repo.all(GroupMembershipRequest) |> IO.inspect
+            conn
+            |> put_flash(:info, "Sent invitation to join group successfully.")
+            |> redirect(to: group_invitation_path(conn, :index, group_id))
+          {:error, _changeset} ->
+            redirect_and_flash_error(conn, "Error inviting to group", group_id)
+        end
+    end
   end
 
   def update(conn, %{"group_membership_request" => _params}) do
     conn
+  end
+
+  defp preloaded_group_or_nil(group_id) do
+    group = case Repo.get(Group, group_id) do
+      nil -> nil
+      group -> Repo.preload(group, :users)
+    end
+  end
+
+  defp redirect_and_flash_error(conn, message, group_id) do
+    conn
+    |> put_flash(:error, message)
+    |> redirect(to: group_invitation_path(conn, :index, group_id))
   end
 end
--- a/web/controllers/group_invitation_controller.ex
+++ b/web/controllers/group_invitation_controller.ex
@@ -23,10 +23,7 @@ defmodule Pairmotron.GroupInvitationController do
   def new(conn, %{"group_id" => group_id}) do
     changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{}, %{})
     group = Repo.get(Group, group_id)
-    invitable_users = User.users_not_in_group(group)
-      |> Repo.all
-      |> Enum.map(&["#{&1.name}": &1.id])
-      |> List.flatten
+    invitable_users = invitable_users_for_select(group)
     render(conn, "new.html", changeset: changeset, group: group, invitable_users: invitable_users)
   end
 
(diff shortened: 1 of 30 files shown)


mbramson:
@ericworkman still trying to work out how to construct an Ecto query that returns all users that aren't in a specific group, but the changeset validates that the user isn't in the group anyhow, so this is functionally complete and ready for review!

web/controllers/group_invitation_controller.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  alias Pairmotron.{Group, GroupMembershipRequest, User, UserGroup}
+  import Pairmotron.ControllerHelpers
+
+  plug :load_resource, model: GroupMembershipRequest, only: [:update]
+
+  def index(conn, %{"group_id" => group_id}) do
+    group = Repo.get(Group, group_id)
+    if group do
+      group = group |> Repo.preload([{:group_membership_requests, :user}])
```
ericworkman: You should move the preload to below the owner check. No reason to make this query if we don't have to.

web/models/user.ex
```diff
@@ -68,4 +69,17 @@ defmodule Pairmotron.User do
     Pairmotron.User
     |> Ecto.Query.where([u], u.active)
   end
+
+  def users_not_in_group(%Pairmotron.Group{} = group), do: users_not_in_group(group.id)
+  def users_not_in_group(group_id) do
+    from user in Pairmotron.User,
+    order_by: user.name
+
```
ericworkman: This'll work:
```
from user in User,
  left_join: user_group in UserGroup, on: user_group.user_id == user.id and user_group.group_id == ^group_id,
  where: is_nil(user_group.user_id)
```

web/controllers/group_invitation_controller.ex
```diff
@@ -0,0 +1,119 @@
...
+        redirect_and_flash_error(conn, "You must be the owner of a group to invite user to that group", group_id)
+      true ->
+        implicit_params = %{"initiated_by_user" => false, "group_id" => group_id}
+        final_params = Map.merge(group_membership_request_params, implicit_params)
+        changeset = GroupMembershipRequest.users_changeset(%GroupMembershipRequest{}, final_params, group)
+
+        case Repo.insert(changeset) do
+          {:ok, _group_membership_request} ->
+            conn
+            |> put_flash(:info, "Sent invitation to join group successfully.")
```
ericworkman: Any chance we can put the group name into this message?

ericworkman: And maybe the invited user's name?

web/controllers/group_invitation_controller.ex
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  defp redirect_and_flash_error(conn, message, group_id) do
+    conn
+    |> put_flash(:error, message)
+    |> redirect(to: group_invitation_path(conn, :index, group_id))
+  end
+
+  defp preloaded_group_or_nil(group_id) do
```
ericworkman: I'm not a fan of thing_or_nil. I'd rather bail out early if the group doesn't exist than return nil and bail a little later.

web/controllers/group_invitation_controller.ex
```diff
@@ -0,0 +1,119 @@
...
+    case Repo.get(Group, group_id) do
+      nil -> nil
+      group -> Repo.preload(group, :users)
+    end
+  end
+
+  defp update_transaction(group_membership_request, user_group_changeset) do
+    Ecto.Multi.new
```
ericworkman: Nice!

web/controllers/users_group_membership_request_controller.ex
```diff
@@ -0,0 +1,89 @@
...
+          {:error, :user_group, %{errors: [user_id_group_id: _]}, _} ->
+            Repo.delete!(group_membership_request)
+            redirect_and_flash_error(conn, "You are already in that group!")
+          {:error, :user_group, _, _} ->
+            redirect_and_flash_error(conn, "Error adding user to group")
+        end
+    end
+  end
+
+  defp user_is_in_group?(user, group_id) do
```
ericworkman, line 72: I think we do this elsewhere. Perhaps we need a helper for this.

web/controllers/users_group_membership_request_controller.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  def create(conn, %{"group_membership_request" => group_membership_request_params}) do
+    current_user = conn.assigns.current_user
+    group_id = parameter_as_integer(group_membership_request_params, "group_id")
+
+    cond do
+      user_is_in_group?(current_user, group_id) ->
+        redirect_and_flash_error(conn, "User already in group")
+      true ->
+        implicit_params = %{"user_id" => current_user.id, "initiated_by_user" => true}
```
ericworkman, line 26: 👍 for calling out implicitness.

ericworkman (changes requested):
Looking solid! The invitation controllers seems a little heavy, but we can refactor those later, and I've added a couple comments here and there. I think the biggest change I'd like to see is a little bit of UX improvement. Invitations are a little hard to find. The no_groups template may be a good place for some of the invite actions to go as well, as that'd be the most actionable thing for a user to do from that template.

web/templates/group_invitation/index.html.eex
```diff
@@ -0,0 +1,41 @@
...
+        </td>
+        <%= if group_membership_request.initiated_by_user do %>
+          <td>User Requested</td>
+          <td>
+            <%= render "accept_invitation_form.html",
+                  conn: @conn,
+                  action: group_invitation_path(@conn, :update, group_membership_request.group_id, group_membership_request) %>
+          </td>
+        <%= else %>
+          <td>Invited by Group</td>
```
ericworkman: I don't think we need to be showing users this checkbox.

web/templates/users_group_membership_request/index.html.eex
```diff
@@ -0,0 +1,34 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <%= for group_membership_request <- @group_membership_requests do %>
+      <tr>
+        <td>
+          <%= link group_membership_request.group.name, to: group_path(@conn, :show, group_membership_request.group) %>
+        </td>
+        <%= if group_membership_request.initiated_by_user do %>
+          <td>User Requested</td>
```
ericworkman: I don't think we need to be showing users this column.

web/controllers/group_invitation_controller.ex
mbramson: Ya that's a good point. I think I was overthinking giving a useful message in the case that a user goes to the route for a non-existent group. Without going directly to an erroneous url, the only way this could happen simply interacting with the UI is if the group is deleted between when someone tries to invite a user and when they click submit. I changed this to a `Repo.get!` so that it explodes instantly.

web/controllers/users_group_membership_request_controller.ex
mbramson: Good call. I'll look for it.

web/models/user.ex
mbramson: Awesome!

web/controllers/users_group_membership_request_controller.ex
mbramson: So the other place that's doing it is in the validate_user_is_in_group? method which is a changeset validation. You're definitely right that it's sort of duplicate code. I'm not sure it's the time to extract yet.

web/templates/group_invitation/index.html.eex
mbramson: Are you saying that we don't need to show users this column? I'm a bit confused by "checkbox".

ericworkman: Ah, yeah, sorry, slip up. Column it is.

ericworkman (approved):

PR #97, opened 2017-02-20
Fix 1-pairs not being 'dead' with new users.

ericworkman:
1-pairs should always be considered 'dead' once new users are available to pair.
1-pairs are a sort of last resort, and they should always be reconfigured if there are better options.

@mbramson This bug most easily manifested when there was 1 existing pair of 1 user and another user became available, as the test case shows. I think its safe to 💀 any existing 1-pairs when repairifying.


mbramson (approved):
LGTM

PR #100, opened 2017-02-22
Create no groups shared template

ericworkman:
@mbramson I hijacked the project no_group template as a shared template, but I wanted to get your feedback on this before I do any more. I'll call out a couple things in comments here in a moment.

As Anna, a user with no current groups:
She has no groups on /pair

screen shot 2017-02-22 at 12 13 22 pm

She has no groups in /profile

screen shot 2017-02-22 at 12 13 30 pm

She has no groups in /projects

screen shot 2017-02-22 at 12 13 40 pm

As Eric, a user in a couple of groups (but turned inactive for sake of showing this):
His /pairs shows his groups are empty. This is what users will see if they look in the past or future

screen shot 2017-02-22 at 12 14 24 pm

His groups show no pairs as well

screen shot 2017-02-22 at 12 14 38 pm

### DIFF
--- a/web/templates/pair/index.html.eex
+++ b/web/templates/pair/index.html.eex
@@ -8,6 +8,12 @@
         <%= for pair <- pairs do %>
           <%= render Pairmotron.SharedView, "pair.html", pair: pair, conn: @conn %>
         <% end %>
+        <%= if Enum.empty?(pairs) do %>
+          <p>No pairs available</p>
+        <% end %>
+      <% end %>
+      <%= if Enum.empty?(@groups_and_pairs) do %>
+        <%= render Pairmotron.SharedView, "no_group.html", conn: @conn %>
       <% end %>
     </ul>
   </div>
--- a/web/templates/project/no_groups.html.eex
+++ b/web/templates/project/no_groups.html.eex
@@ -1,14 +1,1 @@
-<p>
-  <strong>
-    Projects have to be associated with groups. It looks like you aren't in a group. Find a group to join or create your own!
-  </strong>
-</p>
-
-<p>
-  <a href=<%= group_path(@conn, :index) %>>
-    <button class="btn btn-primary">Find Group</button>
-  </a>
-  <a href=<%= group_path(@conn, :new) %>>
-    <button class="btn btn-primary">Create Group</button>
-  </a>
-</p>
+<%= render Pairmotron.SharedView, "no_group.html", conn: @conn %>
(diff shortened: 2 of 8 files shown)


web/templates/pair/index.html.eex
```diff
@@ -8,6 +8,12 @@
         <%= for pair <- pairs do %>
           <%= render Pairmotron.SharedView, "pair.html", pair: pair, conn: @conn %>
         <% end %>
+        <%= if Enum.empty?(pairs) do %>
+          <p>No pairs available</p>
+        <% end %>
+      <% end %>
+      <%= if Enum.empty?(@groups_and_pairs) do %>
+        <%= render Pairmotron.SharedView, "no_group.html", conn: @conn %>
```
ericworkman, line 16: I can use a shared view and render the template as needed

web/templates/project/no_groups.html.eex
```diff
@@ -1,14 +1 @@
...
-
-<p>
-  <a href=<%= group_path(@conn, :index) %>>
-    <button class="btn btn-primary">Find Group</button>
-  </a>
-  <a href=<%= group_path(@conn, :new) %>>
-    <button class="btn btn-primary">Create Group</button>
-  </a>
-</p>
+<%= render Pairmotron.SharedView, "no_group.html", conn: @conn %>
```
ericworkman, line 1: Even here

web/templates/shared/no_group.html.eex
```diff
@@ -0,0 +1,14 @@
+<p>
+  <strong>
+    It looks like you aren't in a group. Find a group to join or create your own!
```
ericworkman, line 3: Not project specific anymore

web/templates/project/no_groups.html.eex
ericworkman: Or I can do what's being done in the projects controller, and render a page like this.

ericworkman (commented):
@mbramson thoughts? Either choose to render in the template or choose a different template to render. Going with the first complicates the templates, but the second way complicates controllers and requires addition duplication of any header-like components if they are similar (see the /projects example vs the /pair and /group/#/pair examples).

web/templates/pair/index.html.eex
mbramson: Yep, I definitely like this.

web/templates/project/no_groups.html.eex
mbramson: This is great. Makes a lot more sense.

web/templates/shared/no_group.html.eex
mbramson: Ya, I definitely agree with the language

mbramson (approved):
LGTM

PR #101, opened 2017-02-24
Use date range instead of week of year

### LINKED ISSUE #30: Change current page display to a date range
Per the conversation in #28, the current method of displaying which week it is:

> Pairs for Week 40 of 2016

runs into some user experience problems at the end of the year for certain years. Users might see a year that is actually the current week, but isn't actually the correct year.

A better method of displaying the current week would be something like:

> Pairs for 11/14/2016 to  11/20/2016

Some things to note:
Weeks start on Mondays and end on Sundays.
There should be a new test for a week that wraps around the end of the year.

ericworkman:
Closes #30 
@mbramson 

screen shot 2017-02-24 at 4 48 31 pm


mbramson (approved):
LGTM

PR #105, opened 2017-02-28
Extract login setup for controller tests

mbramson:
There was a lot of duplicate code in the controller tests building the connection, the user, and logging in the user.

This PR extracts that to `Pairmotron.LoginHelper` which is automatically imported via `use Pairmotron.ConnCase`.

Controller Tests have also had their actions separated out by describes to allow for shortened test descriptions and easier future test implementation.


ericworkman (approved):
LGTM

PR #108, opened 2017-02-28
Remove login link at bottom of register form

austenmadden:
Small UX change to the register form.

## Why

It's easy to confuse at a glance for the form submit for the register form. When it is clicked it drops all of the entries and the user needs to restart. It's in the nav bar so I don't see why it needs to also be at the bottom of the register form.

## How

Remove it.

### DIFF
--- a/web/templates/registration/new.html.eex
+++ b/web/templates/registration/new.html.eex
@@ -31,7 +31,6 @@
   </div>
 
   <div class="form-group">
-    <%= link("Login", to: session_path(@conn, :new), class: "btn btn-success") %>
     <%= submit "Register", class: "btn btn-primary pull-right" %>
   </div>
 <% end %>


Serneum (approved):
LGTM

mbramson (approved):
LGTM

ericworkman (approved):

austenmadden:
@ericworkman @Serneum @mbramson Wonder if it's a result of this work coming from a fork? Only thing different from the other pr that CI had passing

ericworkman:
@austenmadden That looks likely. We have an env var for PG_USER setup in Travis. I did not realize that forks don't get to use those env vars, as they are treated as encrypted secrets and not given to fork builds.

If @mbramson and @Serneum are cool with this, this can be merged. Afterwards, we'll move the env var into the Travis config file in the repo.

Serneum:
Merge it. If it fails, figure it out then

ericworkman:
@austenmadden Thank you!

PR #109, opened 2017-02-28
Make registration use User.registration_changeset

### LINKED ISSUE #106: Display errors in sign up form
When I initially registered for parimotron using my email address `scottmascio@gmail.com` with password `foo_bar` I was under the impression that my account was created because no errors were thrown.

When I went to sign in I was unable too. Turns out my account wasnt created.

I am not sure why it wasnt created, I assume an error was thrown on the registration page but no error was displayed.

mbramson:
registration was erroneously using User.profile_changeset, which as of a
recent fix allowed users to hit the non-null password db constraint when
registering rather than seeing a helpful error message around the
password field stating that it can't be blank

closes #106

### DIFF
--- a/web/controllers/registration_controller.ex
+++ b/web/controllers/registration_controller.ex
@@ -10,7 +10,7 @@ defmodule Pairmotron.RegistrationController do
   end
 
   def create(conn, %{"user" => user_params}) do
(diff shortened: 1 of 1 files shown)


ericworkman (approved):
LGTM, wait until Heroku is up before merging

PR #110, opened 2017-02-28
Add description to groups

ericworkman:
@mbramson  Screenshots incoming when I can.


ericworkman:
@mbramson ready

test/controllers/group_controller_test.exs
```diff
@@ -18,7 +18,86 @@ defmodule Pairmotron.GroupControllerTest do
...
       conn = get conn, group_path(conn, :index)
-      assert html_response(conn, 200) =~ "Listing groups"
+      assert html_response(conn, 200) =~ "Groups"
+    end
+
+    test "does not show invitations link if user is not group owner", %{conn: conn} do
+      group = insert(:group)
+      conn = get conn, group_path(conn, :index)
+      refute html_response(conn, 200) =~ group_invitation_path(conn, :index, group)
+    end
```
mbramson, line 28: What if the user is in the group but not the owner?

web/controllers/group_controller.ex
```diff
@@ -7,10 +7,20 @@ defmodule Pairmotron.GroupController do
...
   end
 
+  def show(conn, %{"id" => id}) do
+    group = Repo.get!(Group, id) |> Repo.preload(:owner)
+    invite_changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{}, %{group_id: id})
+    current_user = conn.assigns.current_user |> Repo.preload([:groups, :group_membership_requests])
+    conn = conn |> Plug.Conn.assign(:current_user, current_user)
+    render(conn, "show.html", group: group, invite_changeset: invite_changeset)
+  end
+
```
mbramson: Why did you move the show to before the :new action? Just curious, the default ordering that phoenix spawns for you has :show between the :create and :edit, which I think kind of makes sense since it can only be used on a resource that occurs after the :create action has been used.

web/templates/group/group_actions.html.eex
```diff
@@ -0,0 +1,24 @@
...
+          conn: @conn,
+          action: users_group_membership_request_path(@conn, :update, current_user_group_membership_request_for_group(@conn, @group)) %>
+  <%= current_user_in_group?(@conn, @group) -> %>
+    <%= link "Pairs", to: group_pair_path(@conn, :show, @group), class: "btn btn-default btn-sm"%>
+  <%= not current_user_in_group?(@conn, @group) and not current_user_has_requested_membership_to_group?(@conn, @group) -> %>
+    <%= render "request_membership_form.html",
+    changeset: Pairmotron.GroupMembershipRequest.changeset(%Pairmotron.GroupMembershipRequest{}, %{group_id: @group.id}),
+          action: users_group_membership_request_path(@conn, :create) %>
+  <%= true -> %>
+<%= end %>
```
mbramson, line 14: Ah didn't realize you could just leave off the return value here. Nice!

web/views/group_view.ex
```diff
@@ -25,4 +25,14 @@ defmodule Pairmotron.GroupView do
...
+
+  def truncate(nil, _), do: nil
+  def truncate(string, len) do
+    cond do
+      String.length(string) > len ->
+        String.slice(string, 0, len) <> "..."
+      True ->
+        string
+    end
+  end
```
mbramson, line 37: Is 110 sufficient if the string is just `w`'s?

mbramson (commented):
Some thoughts, and a few possible tweaks, but looks awesome!

web/views/group_view.ex
ericworkman: Good call on this. I've added an overflow: scroll property to those boxes. It looks like this, with a scroll bar on the bottom if you hover over it:

screen shot 2017-03-02 at 2 35 51 pm

screen shot 2017-03-02 at 2 36 06 pm

Normally, I think truncating to 110 will catch most cases of this. I'm up for completely different solutions for this, but we should definitely limit this in one way or another.

test/controllers/group_controller_test.exs
ericworkman: Added tests for this. 👍

ericworkman:
@mbramson The CI job timed out, so I kicked it off again.

web/views/group_view.ex
mbramson: Nice! I think that's probably fine for now. Like you said, this shouldn't be an issue for now, and it sounds like a nuanced problem of its own trying to figure out where to truncate based on the width of the characters within.

mbramson (approved):
Great work! LGTM

PR #111, opened 2017-03-01
Fix updating retros with new users or pairs

### LINKED ISSUE #53: Fix being able to create a retrospective for a pair you are not a part of
Currently, you can post to the pair_retro create route to create a retrospective for a pair that you are not a part of. This should result in an error.

The change should occur in the pair_retro changeset function. More information may need to be given to the changeset for this validation to occur.

mbramson:
Fixes manually being able to hit the :update route to update the user or
pair of a retrospective by removing that from the cast parameters
available in an update action to the changeset.

Fix being able to update the pair_date to a date before the date of a
pair if the pair_id parameter was not explicitly supplied.

Add pair and retro ex machina factories

Refactor pair_retro controller tests to use more ex machina and divide
out the actions into their own describe blocks

Disallow creation of retrospectives for pairs that the logged in user is not apart of (through both :new and :create actions).

Closes #53

### DIFF
--- a/test/controllers/pair_retro_controller_test.exs
+++ b/test/controllers/pair_retro_controller_test.exs
@@ -3,7 +3,7 @@ defmodule Pairmotron.PairRetroControllerTest do
 
   alias Pairmotron.PairRetro
   import Pairmotron.TestHelper,
-    only: [log_in: 2, create_pair: 4, create_retro: 2, create_pair_and_retro: 2]
+    only: [create_retro: 2, create_pair_and_retro: 2]
 
   @valid_attrs %{subject: "some content", reflection: "some content", pair_date: Timex.today}
   @invalid_attrs %{}
@@ -16,17 +16,14 @@ defmodule Pairmotron.PairRetroControllerTest do
   defp create_user_and_pair_and_retro() do
     user = insert(:user)
     group = insert(:group, %{owner: user})
-    pair = Pairmotron.TestHelper.create_pair([user], group)
-    retro = create_retro(user, pair)
+    pair = insert(:pair, %{group: group, users: [user]})
+    retro = insert(:retro, user: user, pair: pair)
     {user, pair, retro}
   end
 
-  describe "while authenticated" do
+  describe "using :index while authenticated" do
     setup do
-      user = insert(:user)
-      group = insert(:group, %{owner: user})
-      conn = build_conn() |> log_in(user)
-      {:ok, [conn: conn, logged_in_user: user, group: group]}
+      login_user()
     end
 
     test "lists all entries on index", %{conn: conn} do
@@ -34,7 +31,8 @@ defmodule Pairmotron.PairRetroControllerTest do
       assert html_response(conn, 200) =~ "Listing your retrospectives"
     end
 
-    test "lists the current user's retrospectives", %{conn: conn, logged_in_user: user, group: group} do
+    test "lists the current user's retrospectives", %{conn: conn, logged_in_user: user} do
+      group = insert(:group, %{owner: user, users: [user]})
       {_pair, retro} = create_pair_and_retro(user, group)
       conn = get conn, pair_retro_path(conn, :index)
       assert html_response(conn, 200) =~ Ecto.Date.to_string(retro.pair_date)
@@ -45,15 +43,29 @@ defmodule Pairmotron.PairRetroControllerTest do
       conn = get conn, pair_retro_path(conn, :index)
       refute html_response(conn, 200) =~ Ecto.Date.to_string(retro.pair_date)
     end
+  end
 
-    test "renders form for new resources", %{conn: conn, logged_in_user: user, group: group} do
-      pair = Pairmotron.TestHelper.create_pair([user], group)
+  describe "using :new while authenticated" do
+    setup do
+      login_user()
+    end
+
+    test "renders form for new resources", %{conn: conn, logged_in_user: user} do
+      group = insert(:group, %{owner: user, users: [user]})
+      pair = insert(:pair, %{group: group, users: [user]})
       conn = get conn, pair_retro_path(conn, :new, pair.id)
       assert html_response(conn, 200) =~ "New retrospective"
     end
+  end
 
-    test "creates retro and redirects when data is valid and user is logged in", %{conn: conn, logged_in_user: user, group: group} do
-      pair = Pairmotron.TestHelper.create_pair([user], group)
+  describe "using :create while authenticated" do
+    setup do
+      login_user()
+    end
+
+    test "creates retro and redirects when data is valid and user is logged in", %{conn: conn, logged_in_user: user} do
+      group = insert(:group, %{owner: user, users: [user]})
+      pair = insert(:pair, %{group: group, users: [user]})
       attrs = Map.merge(@valid_attrs, %{pair_id: Integer.to_string(pair.id),
                                         user_id: Integer.to_string(user.id)})
       conn = post conn, pair_retro_path(conn, :create), pair_retro: attrs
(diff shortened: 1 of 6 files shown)


web/models/pair.ex
```diff
@@ -4,16 +4,16 @@ defmodule Pairmotron.Pair do
   schema "pairs" do
     field :year, :integer
     field :week, :integer
-    many_to_many :users, Pairmotron.User, join_through: "users_pairs"
+    many_to_many :users, Pairmotron.User, join_through: Pairmotron.UserPair
```
mbramson, line 7: Had to make this explicit so that the ex machina factory wouldn't encounter a Postgrex error because timestamps were not inserted.

test/controllers/pair_retro_controller_test.exs
```diff
@@ -15,26 +14,24 @@ defmodule Pairmotron.PairRetroControllerTest do
...
   end
 
-  describe "while authenticated" do
+  describe "using :index while authenticated" do
     setup do
-      user = insert(:user)
-      group = insert(:group, %{owner: user})
-      conn = build_conn() |> log_in(user)
-      {:ok, [conn: conn, logged_in_user: user, group: group]}
+      login_user()
```
mbramson, line 25: I ended up extracting all group insertions from setup blocks. Not all tests actually needed a group, so it was an extra db insert in many cases.

test/controllers/pair_retro_controller_test.exs
```diff
@@ -2,8 +2,7 @@ defmodule Pairmotron.PairRetroControllerTest do
   use Pairmotron.ConnCase
 
   alias Pairmotron.PairRetro
-  import Pairmotron.TestHelper,
-    only: [log_in: 2, create_pair: 4, create_retro: 2, create_pair_and_retro: 2]
+  import Pairmotron.TestHelper, only: [create_pair_and_retro: 2]
```
mbramson, line 5: Managed to replace all usages of those methods with ex machina calls.

web/controllers/pair_retro_controller.ex
```diff
@@ -13,28 +13,48 @@ defmodule Pairmotron.PairRetroController do
   end
 
   def new(conn, %{"pair_id" => pair_id}) do
-    conn = assign(conn, :projects, Repo.all(Project))
-    current_user = conn.assigns[:current_user]
-    changeset = PairRetro.changeset(%PairRetro{}, %{pair_id: pair_id, user_id: current_user.id}, nil)
-    render(conn, "new.html", changeset: changeset)
+    current_user = conn.assigns.current_user
+
+    pair = Pair.pair_with_users(pair_id) |> Repo.one
```
mbramson, line 18: So because this uses a query and `Repo.one` instead of `Repo.get |> Repo.preload`, not only is it only one database call, but it also elegantly handles when no record exists.

query `|> Repo.one` returns nil if no record is found, while `Repo.get |> Repo.preload` throws an exception if no record is found. This lets us handle it with the `is_nil(pair)` case very cleanly.

web/controllers/pair_retro_controller.ex
```diff
@@ -13,28 +13,48 @@ defmodule Pairmotron.PairRetroController do
...
+
+    cond do
+      is_nil(pair) ->
+        redirect_and_flash_error(conn, "You cannot create a retrospective for non-existent pair")
+      not current_user.id in Enum.map(pair.users, &(&1.id)) ->
+        redirect_and_flash_error(conn, "You cannot create a retrospective for a pair you are not in")
+      true ->
+        implicit_params = %{"user_id" => conn.assigns.current_user.id}
+        final_params = pair_retro_params |> Map.merge(implicit_params)
```
mbramson, line 46: Decided to make the user implicitly the logged in user. It's safer, results in cleaner code, and the argument against doing so was largely based on giving admins the ability to interact with retros before ex_admin was added.

web/controllers/pair_retro_controller.ex
```diff
@@ -92,4 +114,10 @@ defmodule Pairmotron.PairRetroController do
       pair -> Pairmotron.Calendar.first_date_of_week(pair.year, pair.week)
     end
   end
+
+  defp redirect_and_flash_error(conn, message) do
+    conn
+    |> put_flash(:error, message)
+    |> redirect(to: pair_path(conn, :index))
```
mbramson, line 121: In the case of an error, it simply redirects to /pairs. Does that sound right?

web/models/pair_retro.ex
```diff
@@ -34,6 +34,28 @@ defmodule Pairmotron.PairRetro do
...
+  The update changeset does not allow the user to change the user or pair
+  associated with the pair_retro.
+  """
+  def update_changeset(struct, params \\ %{}, pair_start_date) do
+    struct
+    |> cast(params, @required_update_fields, @optional_fields)
+    |> foreign_key_constraint(:project_id)
+    |> validate_field_is_not_before_date(:pair_date, pair_start_date)
+    |> validate_field_is_not_in_future(:pair_date)
+  end
```
mbramson, line 57: Completely removed the ability to change the associated user or pair at all once a retro exists. Don't have to validate, and you can't do it from the UI anyhow.

mbramson (commented):
Commentary and thoughts

web/models/pair.ex
```diff
@@ -4,21 +4,34 @@ defmodule Pairmotron.Pair do
...
     struct
     |> cast(params, @required_fields, @optional_fields)
   end
+
+  @doc """
+  Ecto query that returns a pair with its :users association prelodaded.
+  Performs a single database call.
+  """
+  def pair_with_users(pair_id) when is_binary(pair_id), do:
+    Integer.parse(pair_id) |> elem(0) |> pair_with_users
```
ericworkman, line 30: 👍

web/controllers/pair_retro_controller.ex
```diff
@@ -13,28 +13,48 @@ defmodule Pairmotron.PairRetroController do
...
-    current_user = conn.assigns[:current_user]
-    changeset = PairRetro.changeset(%PairRetro{}, %{pair_id: pair_id, user_id: current_user.id}, nil)
-    render(conn, "new.html", changeset: changeset)
+    current_user = conn.assigns.current_user
+
+    pair = Pair.pair_with_users(pair_id) |> Repo.one
+
+    cond do
+      is_nil(pair) ->
+        redirect_and_flash_error(conn, "You cannot create a retrospective for non-existent pair")
```
ericworkman: `for a non-existent`

web/controllers/pair_retro_controller.ex
```diff
@@ -13,28 +13,48 @@ defmodule Pairmotron.PairRetroController do
...
-    else
-      redirect_not_authorized(conn, pair_retro_path(conn, :index))
+    current_user = conn.assigns.current_user
+
+    pair_id = parameter_as_integer(pair_retro_params, "pair_id")
+    pair = Pair.pair_with_users(pair_id) |> Repo.one
+
+    cond do
+      is_nil(pair) ->
+        redirect_and_flash_error(conn, "You cannot create a retrospective for non-existent pair")
```
ericworkman: `for a non-existent`

web/controllers/pair_retro_controller.ex
ericworkman: That's fine. `/pairs/<year>/<week>` would work slightly better.

ericworkman (changes requested):

web/controllers/pair_retro_controller.ex
mbramson: Unfortunately, we don't reliable have access to the year and week here. I could do a case statement in the redirect_and_flash_error if you like that redirects differently if the week and year are available?

ericworkman: I'm fine with how it is. 👍

PR #113, opened 2017-03-03
Improve admin usage of /groups

ericworkman:
@mbramson This fixes a few UX issue with groups if you are an admin user. Mostly added some spacing around the form elements in the group boxes, as admins have more buttons that normal. Also noted some odd behavior as admin for labels that I missed before.

### DIFF
--- a/test/controllers/group_controller_test.exs
+++ b/test/controllers/group_controller_test.exs
@@ -36,7 +36,7 @@ defmodule Pairmotron.GroupControllerTest do
     test "does not show edit group link if user is not group owner", %{conn: conn} do
       group = insert(:group)
       conn = get conn, group_path(conn, :index)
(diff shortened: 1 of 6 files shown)


mbramson:
Did you intend to have this border on the right and below a group?
screen shot 2017-03-03 at 10 55 10 am

The lower right corner is a different shade as well.

ericworkman:
That's interesting. This is what it looks like on both Firefox and Chrome
screen shot 2017-03-03 at 11 09 36 am

What browser are you using?

mbramson:
Internet Explorer... I jest, I was using Chrome

mbramson:
So figured I'd check what this looks like in prod, and I have those same boxes there as well. So this is probably a me issue, or at the very least, this PR won't change anything about it.

LGTM!

mbramson (approved):
LGTM

PR #115, opened 2017-03-04
Replace projects conn assign with parameter

### LINKED ISSUE #58: Make PairRetroController :new and :edit actions pass projects parameter
Currently, the PairRetroController assigns the list of all projects to the conn.assigns where it is retrieved through a method on PairRetroView and then passed to the form in the web/templates/pair_retro/new.html.eex and web/templates/pair_retro/edit.html.eex. 

This should probably be refactored so that the PairRetroController passes the projects as a parameter directly to the new.html.eex and edit.html.eex templates. This eliminates the View method, and generally cleans up the code.

### LINKED ISSUE #84: Properly handle the update of a retrospective's associated project
There are some edge cases associated with updating a retrospective's associated project.

The list should only contain the projects associated with the group that the pair is associated with.

If the user has been removed from a group and edits an existing retrospective associated with a pair (associated with that group) they had when they were still in that group, they should not be allowed to change the project.

They should be able to edit the project field if they are still in the group associated with the retrospective (through the pair).

mbramson:
The PairRetro :new and :edit actions were assigning the projects used
for the dropdown to the conn rather than simply passing through the
projects to the view.

This also fixes a bug where if any errors occurred creating or updating
a PairRetro, the projects dropdown would not populate.

Refactor PairRetroController and the PairRetro model.
  - Add pair and project parameters to the changesets.
  - Remove earliest_pair_date paramter from changeset (since this information is on the pair that is now passed in).
  - Form for retros now only lists projects that are associated with the retro through the retro's pair's group.

Closes #58

Closes #84 

Further Issues
- [x] Disallow :edit and :update actions if the user is no longer in the group associated with the retro's pair. Also make the edit button the PairRetro :index not appear for retros where this is the case.

### DIFF
--- a/web/controllers/pair_retro_controller.ex
+++ b/web/controllers/pair_retro_controller.ex
@@ -71,8 +71,8 @@ defmodule Pairmotron.PairRetroController do
   end
 
   def edit(conn = @authorized_conn, _params) do
-    projects = Repo.all(Project)
-    retro = conn.assigns.pair_retro
+    retro = conn.assigns.pair_retro |> Repo.preload(:pair)
+    projects = Project.projects_for_group(retro.pair.group_id) |> Repo.all
     changeset = PairRetro.changeset(retro, %{}, nil, nil, nil)
     render(conn, "edit.html", pair_retro: retro, changeset: changeset, projects: projects)
   end
--- a/web/controllers/pair_retro_controller.ex
+++ b/web/controllers/pair_retro_controller.ex
@@ -56,7 +56,7 @@ defmodule Pairmotron.PairRetroController do
             |> put_flash(:info, "Pair retro created successfully.")
             |> redirect(to: pair_retro_path(conn, :index))
           {:error, changeset} ->
-            projects = Repo.all(Project)
+            projects = Project.projects_for_group(pair.group_id) |> Repo.all
             render(conn, "new.html", changeset: changeset, projects: projects)
         end
     end
--- a/web/controllers/pair_retro_controller.ex
+++ b/web/controllers/pair_retro_controller.ex
@@ -102,8 +102,10 @@ defmodule Pairmotron.PairRetroController do
   end
 
   defp project_from_params_or_pair_retro(params, pair_retro) do
-    id = Map.get(params, "project_id") || (pair_retro.project && pair_retro.project.id) || 0
-    Repo.get(Project, id)
+    case Map.get(params, "project_id") || (pair_retro.project && pair_retro.project.id) do
+      nil -> nil
+      id -> Repo.get(Project, id)
+    end
   end
 
   def delete(conn = @authorized_conn, _params) do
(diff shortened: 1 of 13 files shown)


web/models/pair.ex
```diff
@@ -26,12 +26,18 @@ defmodule Pairmotron.Pair do
...
-  def pair_with_users(pair_id) when is_binary(pair_id), do:
-    Integer.parse(pair_id) |> elem(0) |> pair_with_users
   def pair_with_users(pair_id) do
     from pair in Pairmotron.Pair,
     join: users in assoc(pair, :users),
     where: pair.id == ^pair_id,
     preload: [users: users]
   end
+
+  def pair_with_users_and_group(pair_id) do
```
ericworkman: Is this solely used for testing?

web/controllers/pair_retro_controller.ex
```diff
@@ -67,35 +70,42 @@ defmodule Pairmotron.PairRetroController do
...
+        projects = Project.projects_for_group(pair.group_id) |> Repo.all
+        render(conn, "edit.html", pair_retro: pair_retro, changeset: changeset, projects: projects)
     end
   end
   def update(conn, _params) do
     redirect_not_authorized(conn, pair_retro_path(conn, :index))
   end
 
+  defp project_from_params_or_pair_retro(params, pair_retro) do
+    id = Map.get(params, "project_id") || (pair_retro.project && pair_retro.project.id) || 0
```
ericworkman: The `|| 0` is here so that the `Repo.get` works right? I think it would be better to return `nil` instead of making a database call and relying on there never being a Project 0.

web/models/pair_retro.ex
```diff
@@ -48,15 +52,20 @@ defmodule Pairmotron.PairRetro do
...
+    |> validate_date_is_not_before_pair(:pair_date, pair)
     |> validate_field_is_not_in_future(:pair_date)
+    |> validate_project_is_for_group(:project_id, project, pair)
   end
 
-  defp validate_field_is_not_before_date(changeset, field, pair_start_date) do
+  defp validate_date_is_not_before_pair(changeset, field, pair) do
+    pair_start_date = cond do
+      is_nil(pair) -> nil
+      true -> Pairmotron.Calendar.first_date_of_week(pair.year, pair.week)
```
ericworkman: This could be two functions if you pattern match on pair == nil. Looks like it might simplify logic below in the changeset validation.

ericworkman (commented):
Looks good, just a couple comments.

web/controllers/pair_retro_controller.ex
mbramson: Ya, that's exactly why. The trouble is `Repo.get(Project, nil)` errors. There really should never be a project 0, as the index starts at 0 and counts up. There might be a better way to handle this.

web/models/pair_retro.ex
mbramson: Definitely! Good call.

web/models/pair.ex
mbramson: Oooooh. This is vestigial. I forgot to remove it.

mbramson:
@ericworkman addressed your concerns, let me know what you think!

ericworkman (approved):
LGTM!

PR #116, opened 2017-03-05
Wrap db transaction in generate_pairs

### LINKED ISSUE #92: Wrap DB actions for generate_pairs in transaction using Ecto.Multi
When generating pairs, there are a number of database operations that happen sequentially. These should be wrapped in a transaction to prevent any weird issues with partially applied changes. The relvant code is located at https://github.com/ElixirCLE/pairmotron/blob/master/web/controllers/pair_controller.ex#L74-L90.

Ecto.Multi is Ecto's way of utilizing database transactions, and should be used to implement this.

ericworkman:
@mbramson PairMaker.generate_pairs/3 now uses Ecto.Multi and a transaction. Upped the ecto deps while I was at it.

Closes #92


lib/pairmotron/pair_maker.ex
```diff
@@ -47,8 +47,12 @@ defmodule Pairmotron.PairMaker do
 
     determination = PairBuilder.determify(pairs, users)
 
-    determination.dead_pairs
-      |> Enum.map(fn(p) -> Repo.delete! p end)
+    multi = Ecto.Multi.new
+
+    multi = determination.dead_pairs
+      |> Enum.reduce(multi, fn(pair, acc) ->
+          acc |> Ecto.Multi.delete(:remove_dead_pairs, pair)
```
ericworkman: See an alternative way of doing this? Since we have to apply "actions" on a multi, it acts exactly like an accumulator.

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +62,23 @@ defmodule Pairmotron.PairMaker do
...
+    multi |> Ecto.Multi.insert(:create_lone_user_pair, user_pair)
   end
 
-  defp make_pairs(users, year, week, group_id) do
-    pair = Repo.insert! Pair.changeset(%Pair{}, %{year: year, week: week, group_id: group_id})
-    users
-      |> Enum.map(fn(user) -> UserPair.changeset(%UserPair{}, %{pair_id: pair.id, user_id: user.id}) end)
+  defp insert_pair(multi, year, week, group_id, users) do
+    pair = Pair.changeset(%Pair{}, %{year: year, week: week, group_id: group_id})
+      |> Ecto.Changeset.put_assoc(:users, users)
```
ericworkman: This was a lot better than creating %UserPair structs.

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +62,23 @@ defmodule Pairmotron.PairMaker do
       |> Mixer.mixify(week)
       |> Pairer.generate_pairs(pairs)
 
-    results.pairs
-      |> Enum.map(fn(users) -> make_pairs(users, year, week, group_id) end)
-      |> List.flatten
-      |> Enum.map(fn(p) -> Repo.insert! p end)
+    multi = results.pairs
+      |> Enum.reduce(multi, fn(users, acc) ->
+          insert_pair(acc, year, week, group_id, users)
```
ericworkman: Same accumulator behavior.

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +62,23 @@ defmodule Pairmotron.PairMaker do
...
+      end)
 
-    insert_user_pair(results.user_pair)
+    multi = insert_lone_user_pair(multi, results.user_pair)
+    Repo.transaction(multi)
```
ericworkman: I'm not sure if this should care about the success for failure of the transaction. The controller might, but I'm not sure if anything would be useful save a "Something failed during repairifying" message. Thoughts?

lib/pairmotron/pair_maker.ex
```diff
@@ -26,7 +26,7 @@ defmodule Pairmotron.PairMaker do
       |> Repo.all
   end
 
-  defp fetch_users_from_pairs(pairs) do
+  defp preload_users(pairs) do
```
ericworkman: `fetch_users_from_pairs` was a terrible name.

mbramson: Come to think of it, I'm not sure why this was extracted. All it's doing is preloading users. Couldn't we just pipe to `Repo.preload([:users])` directly on line 10?

lib/pairmotron/pair_maker.ex
mbramson: You might be able to use Multi.delete_all as well, but I think you might have to construct a query there. You might gain some clarity. This looks fine though.

lib/pairmotron/pair_maker.ex
mbramson: You're definitely right that the user doesn't really gain much from seeing this fail, but something rubs me the wrong way about this failing with no trace. If this starts failing, I'd want to know about it (because it really shouldn't). Can we at least log in the case of a failure?

lib/pairmotron/pair_maker.ex
mbramson: Oooooh. Ya that is way more elegant!

mbramson (commented):
This looks awesome. Just a few thoughts.

ericworkman:
@mbramson that got a little hairier than I thought it would. Found an error with the multi during more manual testing.

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +69,50 @@ defmodule Pairmotron.PairMaker do
...
 
-    results.pairs
-      |> Enum.map(fn(users) -> make_pairs(users, year, week, group_id) end)
-      |> List.flatten
-      |> Enum.map(fn(p) -> Repo.insert! p end)
+    Ecto.Multi.new
+      |> remove_dead_pairs(determination.dead_pairs)
+      |> insert_pairs(results.pairs, year, week, group_id)
+      |> insert_lone_user_pair(results.user_pair)
+      |> commit_transaction(year, week, group_id)
```
mbramson, line 76: Love how well this reads now

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +69,50 @@ defmodule Pairmotron.PairMaker do
...
+          insert_pair(acc, year, week, group_id, users)
+      end)
   end
 
-  defp make_pairs(users, year, week, group_id) do
-    pair = Repo.insert! Pair.changeset(%Pair{}, %{year: year, week: week, group_id: group_id})
-    users
-      |> Enum.map(fn(user) -> UserPair.changeset(%UserPair{}, %{pair_id: pair.id, user_id: user.id}) end)
+  defp insert_pair(multi, year, week, group_id, users) do
+    atom = String.to_atom("insert_pair_#{year}_#{week}_#{group_id}_#{hd(users).id}")
```
mbramson: Are we sure it's going to be a good idea to dynamically generate atoms like this? Atoms aren't garbage collected and there is a upper limit to the number of atoms that are allowed to exist (I think it's a million). It'd be pretty crazy to run into that limit, but still.

lib/pairmotron/pair_maker.ex
```diff
@@ -58,23 +69,50 @@ defmodule Pairmotron.PairMaker do
...
+    multi |> Ecto.Multi.insert(atom, pair)
   end
 
+  defp commit_transaction(multi, year, week, group_id) do
+    case Repo.transaction(multi) do
+      {:error, failed_operation, failed_value, _} ->
+        Logger.error "Generating pairs failed"
+        Logger.error "Year: #{year}, week: #{week}, group_id: #{group_id}"
+        Logger.error "#{failed_operation}"
+        Logger.error "#{failed_value}"
```
mbramson, line 115: Solid!

web/controllers/group_pair_controller.ex
```diff
@@ -11,7 +11,13 @@ defmodule Pairmotron.GroupPairController do
 
     group = Repo.get(Group, group_id)
     if authorized?(group, conn.assigns.current_user) do
-      pairs = PairMaker.fetch_or_gen(year, week, group_id)
+      pairs = case PairMaker.fetch_or_gen(year, week, group_id) do
+        {:error, pairs, message} ->
+          conn |> put_flash(:error, message)
```
mbramson: So if I'm not mistaken, this isn't actually going to result in the conn having that flash message down below, since it's not actually being returned by the case statement (pairs is). The conn outside of this case will be unaltered. Yay, Immutability!

web/controllers/pair_controller.ex
```diff
@@ -18,25 +18,31 @@ defmodule Pairmotron.PairController do
...
   end
 
-  defp pairs_for_user_groups(user, year, week) do
+  defp pairs_for_user_groups(user, conn, year, week) do
     user.groups
       |> Enum.map(fn(group) ->
-        PairMaker.fetch_or_gen(year, week, group.id)
+        case PairMaker.fetch_or_gen(year, week, group.id) do
+          {:error, pairs, message} ->
+            conn |> put_flash(:error, message)
```
mbramson: Same here. All put_flash is doing is adding to a property on the conn, which isn't actually returning to anywhere (since this case is returning pairs).

mbramson (commented):
Just a few more thoughts. Handling of the transaction errors looks great!

mbramson (approved):
Assuming you aren't cooking up more changes, this LGTM

PR #123, opened 2017-03-08
Remove retro links from /pairs

ericworkman:
@mbramson Had to rearrange some of the shared template to not just repeat everything.


mbramson (approved):
LGTM. This clears up the current issue, but I think we should add a card to get the retro links working on the /pairs route again.

ericworkman:
@mbramson Agreed.

PR #124, opened 2017-03-08
Fix Ecto.Query.CastError when no project is selected for a retrospective

mbramson:
The PairRetroController :create action wasn't properly handling a pair_id of "", which is what the form returns when the default option of the project "(none)" is selected. It now handles this properly.


ericworkman (approved):
👍 LGTM

PR #127, opened 2017-03-09
Use shuffle in mixer

ericworkman:
Since pairs are no longer deleted unless necessary during repairifying,
  we can use the built in shuffle and accept completely random results.

@mbramson I think the chance that shuffling an 8 element list results in the original order to be sufficiently rare. We can increase this or figure out a better test if you disagree.


mbramson:
As discussed, it'd be nice good to inject Enum.shuffle to mixify rather than relying on chance for a test to fail or not.

ericworkman:
@mbramson ready

mbramson (approved):
LGTM

PR #128, opened 2017-03-09
Fix error while removing multiple pairs during a repairify

ericworkman:
@mbramson found this issue during the mixify -> shuffle change. Same lesson as the previous time.


mbramson (approved):
LGTM

PR #130, opened 2017-03-10
Change pair_retro reflection to be :text type

ericworkman:
@mbramson Noticed that the retro reflection text was limited. Found via an error when I tried to write up a longer retro. The default for `:string` type is `varchar(255)`. This modifies pair_retros to use a `:text` type in Postgres, so it is now unlimited.


mbramson (approved):
LGTM. I'm assuming that this doesn't wipe out existing data in the reflection field.

ericworkman:
@mbramson That comment made me test out roll back. Forgot that I'd have to define down differently, so the rollback failed. This fixes it. Data does not get wiped during migration.

mbramson:
Awesome! We probably want to add a card to add validate_length(:field, 255) to a bunch of changesets where the :text type isn't as appropriate

PR #132, opened 2017-03-11
Eliminate all credo issues

mbramson:
Eliminates all remaining credo issues and adds a credo configuration file.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -1,0 +1,129 @@
+# This file contains the configuration for Credo and you are probably reading
+# this after creating it with `mix credo.gen.config`.
+#
+# If you find anything wrong or unclear in this file, please report an
+# issue on GitHub: https://github.com/rrrene/credo/issues
+#
+%{
+  #
+  # You can have as many configs as you like in the `configs:` field.
+  configs: [
+    %{
+      #
+      # Run any config using `mix credo -C <name>`. If no config name is given
+      # "default" is used.
+      name: "default",
+      #
+      # These are the files included in the analysis:
+      files: %{
+        #
+        # You can give explicit globs or simply directories.
+        # In the latter case `**/*.{ex,exs}` will be used.
+        included: ["lib/", "src/", "web/", "apps/"],
+        excluded: [~r"/_build/", ~r"/deps/"]
+      },
+      #
+      # If you create your own checks, you must specify the source files for
+      # them here, so they can be loaded by Credo before running the analysis.
+      requires: [],
+      #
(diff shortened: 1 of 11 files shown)


.credo.exs
```diff
@@ -0,0 +1,129 @@
+# This file contains the configuration for Credo and you are probably reading
```
mbramson, line 1: This file was added via `mix credo gen.config`.

.credo.exs
```diff
@@ -0,0 +1,129 @@
...
+      # to the tuple.
+      #
+      # To disable a check put `false` as second element:
+      #
+      #     {Credo.Check.Design.DuplicatedCode, false}
+      #
+      checks: [
+        {Credo.Check.Consistency.ExceptionNames},
+        {Credo.Check.Consistency.LineEndings},
+        {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
```
mbramson, line 52: This was my only change to the generated configuration. This makes credo not check for inconsistencies in import style. This check enforces that all imports are either

```
alias Pairmotron.User
alias Pairmotron.Pair
```
or 
```
alias Pairmotron.{User, Pair}
```
This check is a bit buggy, as I eliminated most cases of this inconsistency until I came across an issue where it seems to do the actual check after resolving the `__using__` macro. This is particularly an issue in `Pairmotron.Web`.

After some pontification, I'm not sure that I agree with the check in the first place. While consistency is good, sometimes you'd like to alias many things, and they're not all the same type of thing despite being imported from the same module. As an example, `Pairmotron.Types` defines some types used only for TypeSpecs, while `Pairmotron.{User, Pair, Group}` are all Ecto schemas. It's a bit misleading to lump them all together as `Pairmotron.{User, Pair, Group, Types}`.

lib/pairmotron.ex
```diff
@@ -4,6 +4,7 @@ defmodule Pairmotron do
 
   # See http://elixir-lang.org/docs/stable/elixir/Application.html
   # for more information on OTP Applications
+  @spec start(atom(), map()) :: Supervisor.on_start
```
mbramson, line 7: You could definitely make an argument that these didn't need Typespecs, but they don't hurt, and it was fun tracking them down.

mbramson (commented):
Thoughts and commentary.

lib/pairmotron.ex
ericworkman: `Supervisor.on_start` neato

ericworkman (approved):
Awesome! LGMT!

PR #133, opened 2017-03-12
Add dialyzer

mbramson:
Added dialyzer. I was going to couple that with fixing the output of dialyzer, but... there's a lot of it. I need to read up on it more.


ericworkman (approved):
LGTM

PR #136, opened 2017-03-14
Add input sanitizer

ericworkman:
@mbramson I want to get your feedback on this early. I've only done Group so far, but its enough for feedback.


mbramson (approved):
LGTM. Looks awesome!

ericworkman:
@mbramson Updated the other models. See anywhere else I need to do this?

mbramson (approved):
Nope, I didn't see any other fields that need to be sanitized. If we find any others we can always add them. I say merge it! LGTM

PR #138, opened 2017-03-17
Add ability to delete group invitations

### LINKED ISSUE #102: Allow users to delete group invitations
Currently group invitations are only removed when they are accepted. There is no way to deny a membership request or an invite, all a user can do is ignore it forever.

Implement the :delete action for both the `UsersGroupMembershipRequestController` and the `GroupInvitationController`.

Allow the owners of groups to delete `GroupMembershipRequest`s using the :delete action on the `GroupInvitationController` which should be located on the :index action's template.

Allow users to delete their own `GroupMembershipRequest`s using the :delete action on the `UsersGroupMembershipRequestController` which should be located on the :index action's template.

- [x] Implement controller actions
  - [x] :delete action on `GroupInvitationController`
    - [x] A `GroupMembershipRequest` associated with a `Group` should only be deletable by the group's owner.
  - [x] :delete action on `UsersGroupMembershipRequestController`
    - [x] A `GroupMembershipRequest` associated with a `User` should only be deletable by that user.
- [x] Add delete links to the UI
  - [x] Delete links on `GroupInvitationController` `index.html.eex` template.
  - [x] Delete links on `UsersGroupMembershipRequestController` `index.html.eex` template

mbramson:
Adds controller actions and links on the UI to be able to delete invitations if you are the owner of the group associated with the invitation or the user associated with the invitation.

Closes #102

### DIFF
--- a/lib/pairmotron/invite_delete_helper.ex
+++ b/lib/pairmotron/invite_delete_helper.ex
@@ -20,7 +20,7 @@ defmodule Pairmotron.InviteDeleteHelper do
       Repo.delete!(group_membership_request)
       conn
       |> put_flash(:info, "Group Invite deleted successfully.")
-      |> redirect(to: group_invitation_path(conn, :index, group_membership_request.group_id))
+      |> redirect(to: redirect_path)
     else
       redirect_and_flash_error(conn, "You cannot delete that group invitation", redirect_path)
     end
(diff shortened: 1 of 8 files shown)


ericworkman:
Code looks good to me. I think its slightly confusing to a user when their request gets denied, as there's no feedback to the requesting user or the group owner in the reverse case.

mbramson:
Agreed. As discussed, we would need some sort of notifications system, and that's outside of the scope of this card. The user just not being in the group will have to do for now.

ericworkman (approved):

PR #140, opened 2017-03-17
Add scarier group delete alert message

### LINKED ISSUE #119: Add scarier alert message to group delete
Currently when clicking delete on a group, an alert appears asking "Are you sure?". I think this a bit of an understatement as deleting a group is a pretty significant action. Perhaps the message should inform the user that all Pairs and Projects will be orphaned when they take this action.

mbramson:
Closes #119


ericworkman (approved):
LGTM

Issue #74, opened 2017-01-29
Allow group owners to remove users from their groups


mbramson:
Group owners should be able to remove specific members from their group. The users in a group should be added to the group edit action. In this listing/table, a button/link should be present allowing the removal of that user from the group. This would be a deletion of the UserGroup record.

If the remove action is selected, a confirmation dialog should appear.

mbramson:
Looks like this has already been implemented.

PR #141, opened 2017-03-17
Add UserGroupController allowing users to leave groups

mbramson:
Adds the ability for users to leave a group and also the UI to do so. Technically adds the ability for group owners to be able to delete users from a group but there is currently no UI with which to do so. That will be handled by #74 

Adds the user_group ex_machina factory.

- [x] Adds the UserGroupController and the delete action
- [x] Adds the UI to allow users to leave a group that they are in.
- [x] Add tests around non-existent user or group for deletion
- [x] Add tests around new query in UserGroup.

newgroups


ericworkman (approved):
👍

PR #142, opened 2017-03-18
Fix Ecto.Changeset.cast/4 warnings

mbramson:
Converted all changesets to using Ecto.Changeset.cast/3 and Ecto.Changeset.validate_required/2 instead of Ecto.Changeset.cast/4 which was deprecated in Ecto 2.1.


ericworkman (approved):
👍 Its pretty annoying that `validate_required` takes a list of atoms now. LGTM

PR #143, opened 2017-03-20
Add a list of users to the Group show action

### LINKED ISSUE #114: Add ability to see which users are in a group
Any user should be able to see which users are in a public group.

Only users that are in a specific group should be able to see which users are in that group.

This list should be accessible via a group's show route.

mbramson:
This PR is pretty simple, but lays the groundwork for a lot of other work such as group roles and ejecting users from groups. It's also useful information to know. The only piece of information listed for each user is currently their name.

An alternative implementation would be to add this information to the UserGroup controller under the :index action where it thematically makes sense, and is probably a bit more RESTful. Let me know what you think of this. The reason that I didn't go with this route in the first place, was that I didn't want users to have to travel too far to get to this information (a weak argument).

Closes #114

### DIFF
--- a/web/templates/group/show.html.eex
+++ b/web/templates/group/show.html.eex
@@ -13,7 +13,7 @@
 </h2>
 <p><%= @group.description %></p>
 <%= render "group_actions.html", conn: @conn, group: @group %>
-<h2>Users</h2>
+<h2>Members</h2>
 <table class="table">
   <thead>
     <tr>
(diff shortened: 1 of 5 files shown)


web/templates/group/show.html.eex
```diff
@@ -13,3 +13,21 @@
 </h2>
 <p><%= @group.description %></p>
 <%= render "group_actions.html", conn: @conn, group: @group %>
+<h2>Users</h2>
```
ericworkman: The only change I'd recommend is "Members" instead of "Users"

mbramson:
Good call on "Members". Made the change.

ericworkman (approved):

Issue #126, opened 2017-03-09
Add Create Retrospective / View Retrospective on /pairs index.


ericworkman:
Currently, creating a retrospective can only occur on the group pairs page. I'd like for retros to be available also on the /pairs for each group that a user has a pair. These retros should be unique to each pair (and therefore each group per period). I removed this feature in #123, but would like to see retros return to /pairs.

PR #146, opened 2017-03-21
Refactor pairs controller

mbramson:
Started work on #126, and realized that the PairController was probably in need of a refactor pass.

I think this reduces complexity and makes this section of code a lot easier to reason about.

My favorite part is eliminating the passing down of the conn outside of the index, in favor of passing up a list of messages which are then applied to the flash of the conn in the index.

### DIFF
--- a/web/controllers/pair_controller.ex
+++ b/web/controllers/pair_controller.ex
@@ -12,7 +12,6 @@ defmodule Pairmotron.PairController do
     {year, week} = Timex.iso_week(Timex.today)
     user = conn.assigns[:current_user]
     {conn, groups_and_pairs} = conn
-      |> assign_current_user_pair_retro_for_week(year, week)
       |> fetch_pairs(year, week, user)
     render conn, "index.html", year: year, week: week, groups_and_pairs: groups_and_pairs,
       start_date: Timex.from_iso_triplet({year, week, 1}),
@@ -25,7 +24,6 @@ defmodule Pairmotron.PairController do
     {week, _} = w |> Integer.parse
     user = conn.assigns[:current_user]
     {conn, groups_and_pairs} = conn
-      |> assign_current_user_pair_retro_for_week(year, week)
       |> fetch_pairs(year, week, user)
     render conn, "index.html", year: year, week: week, groups_and_pairs: groups_and_pairs,
       start_date: Timex.from_iso_triplet({year, week, 1}),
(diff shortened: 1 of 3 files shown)


web/controllers/pair_controller.ex
```diff
@@ -4,64 +4,71 @@ defmodule Pairmotron.PairController do
...
-    user = user
-      |> Repo.preload(:groups)
-    {conn, pairs} = user
-      |> pairs_for_user_groups(conn, year, week)
-    {conn, Enum.zip(user.groups, pairs)}
+  @spec fetch_groups_and_pairs(Types.user, integer(), 1..53) :: {[{Types.group, Types.pair}], [String.t]}
+  defp fetch_groups_and_pairs(user, year, week) do
+    user = user |> Repo.preload(:groups)
+    {pairs, messages} = pairs_for_user_groups(user, year, week)
+    {Enum.zip(user.groups, pairs), messages}
```
mbramson, line 43: So I think this line is dangerous. It usually works, but there's no reason to assume that there will always be one pair per group the user is in. It's not crazy to think that there could be something down the line which allows multiple concurrent pairs for one person. Also, as soon as group-specific activity is enabled, this could easily match up pairs with incorrect groups.

I'm not going to fix it in this PR, but I probably will for the PR that fixes #126

mbramson (commented):
A thought

web/controllers/pair_controller.ex
```diff
@@ -4,64 +4,71 @@ defmodule Pairmotron.PairController do
...
-        {conn, result_pairs} = conn_and_pairs
+  @spec pairs_for_user_groups(Types.user, integer(), 1..53) :: {[Types.pair], [String.t]}
+  defp pairs_for_user_groups(user, year, week) do
+    Enum.reduce(user.groups, {[], []},
+      fn(group, pairs_and_messages) ->
+        {result_pairs, messages} = pairs_and_messages
         case PairMaker.fetch_or_gen(year, week, group.id) do
           {:error, pairs, message} ->
-            {conn |> put_flash(:error, message), result_pairs ++ [pairs |> pairs_containing_user(user)]}
+            {result_pairs ++ [pairs |> pairs_containing_user(user)], [message | messages]}
```
ericworkman, line 53: Nice

ericworkman (approved):
LGTM I like the duplication reduction and the restructuring

PR #147, opened 2017-03-23
Add create retro and edit retro links to /pairs

### LINKED ISSUE #126: Add Create Retrospective / View Retrospective on /pairs index.

mbramson:
Added back the create retro and edit retro links to /pairs as they were removed because of a bug when group-based pairing was added.

Previously, the query that retrieved retros simply looked for a retro (and unfortunately multiple) that was associated with the user and the year and week (through a pair). This worked fine when it was only possible for a user to be in one pair per week, but it broke down once a user could be in many pairs in the same week due to being in multiple groups. The end effect was if you were paired in Group A and Group B, and created a retrospective for Group A, the "create retro" link would change to "edit retro" for both Group A and Group B, and would link to the retro that you created for Group A under both links. This now works properly.

This also fixed a subtle bug due to zipping the user's groups with the user's pairs for a given week when a user wasn't in a pair for all of their groups that week. If that pair that they weren't in was not the last group in the list of groups (order was arbitrary), then the pairs could shift to a different group. Pairs are now connected to groups explicitly, so it is possible for the second group out of three to not have a pair for the current user and have it display properly.

Although not a possible case presently, the /pairs route will now properly display multiple pairs for a group if the user is in multiple pairs.

Closes #126

### DIFF
--- a/web/controllers/pair_controller.ex
+++ b/web/controllers/pair_controller.ex
@@ -6,6 +6,8 @@ defmodule Pairmotron.PairController do
 
   alias Pairmotron.PairMaker
 
+  @typep pair_session :: [{Types.group, [{Type.pair, Types.retro | nil}]}]
+
   @spec index(Plug.Conn.t, map()) :: Plug.Conn.t
   def index(conn, _params) do
     {year, week} = Timex.iso_week(Timex.today)
@@ -29,20 +31,20 @@ defmodule Pairmotron.PairController do
     |> render_index(year, week, groups_and_pairs)
   end
 
-  @spec render_index(Plug.Conn.t, integer(), 1..53, [{Types.group, [{Types.pair, Types.retro | nil}]}]) :: Plug.Conn.t
+  @spec render_index(Plug.Conn.t, integer(), 1..53, pair_session) :: Plug.Conn.t
   defp render_index(conn, year, week, groups_and_pairs) do
     render(conn, "index.html", year: year, week: week, groups_and_pairs: groups_and_pairs,
               start_date: Timex.from_iso_triplet({year, week, 1}),
               stop_date: Timex.from_iso_triplet({year, week, 7}))
   end
 
-  @spec fetch_groups_and_pairs(Types.user, integer(), 1..53) :: {[{Types.group, [{Types.pair, Types.retro | nil}]}], [String.t]}
+  @spec fetch_groups_and_pairs(Types.user, integer(), 1..53) :: {pair_session, [String.t]}
   defp fetch_groups_and_pairs(user, year, week) do
     user = user |> Repo.preload(:groups)
     groups_with_pairs_for_user(user, year, week)
   end
 
-  @spec groups_with_pairs_for_user(Types.user, integer(), 1..53) :: {[{Types.group, [{Types.pair, Types.retro | nil}]}], [String.t]}
+  @spec groups_with_pairs_for_user(Types.user, integer(), 1..53) :: {pair_session, [String.t]}
   defp groups_with_pairs_for_user(user, year, week) do
     Enum.reduce(user.groups, {[], []},
       fn(group, accum) ->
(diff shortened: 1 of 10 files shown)


web/controllers/pair_controller.ex
```diff
@@ -29,30 +29,37 @@ defmodule Pairmotron.PairController do
...
   defp fetch_groups_and_pairs(user, year, week) do
     user = user |> Repo.preload(:groups)
-    {pairs, messages} = pairs_for_user_groups(user, year, week)
-    {Enum.zip(user.groups, pairs), messages}
+    groups_with_pairs_for_user(user, year, week)
   end
 
-  @spec pairs_for_user_groups(Types.user, integer(), 1..53) :: {[Types.pair], [String.t]}
-  defp pairs_for_user_groups(user, year, week) do
+  @spec groups_with_pairs_for_user(Types.user, integer(), 1..53) :: {[{Types.group, [{Types.pair, Types.retro | nil}]}], [String.t]}
```
mbramson: So the data structure that we're returning got a bit gnarly. Every group has a list of tuples where that tuple is a pair and possibly a retro (if it exists for that pair and the current user). Definitely open to suggestions on how to make this less mind bending to think about.

web/controllers/pair_controller.ex
```diff
@@ -29,30 +29,37 @@ defmodule Pairmotron.PairController do
...
+            pairs_and_retros = pairs
+              |> pairs_containing_user(user)
+              |> add_retros_to_pairs_for_user(user)
+            new_groups_with_pairs = groups_with_pairs ++ [{group, pairs_and_retros}]
+            {new_groups_with_pairs, messages}
+          {:error, pairs, message} ->
+            pairs_and_retros = pairs
+              |> pairs_containing_user(user)
+              |> add_retros_to_pairs_for_user(user)
+            new_groups_with_pairs = groups_with_pairs ++ [{group, pairs_and_retros}]
```
mbramson: There's definitely duplication between these two cases. I should probably extract that.

mbramson (commented):
Thoughts and commentary

web/controllers/pair_controller.ex
ericworkman: Perhaps its time for a struct for a `PairSession`, which has users and a retro and group attached to it? That way, you are just passing around a list of PairSessions?

web/controllers/pair_controller.ex
ericworkman: Yeah, these are identical except for the message handling, right? Would it be better to adjust `fetch_or_gen` to be consistent about how many items it returns in the :ok and :error tuples?

mbramson: That would make sense actually. In the :ok case, it could just return an empty list of messages, that way duplication goes away without any extraction.

web/controllers/pair_controller.ex
mbramson: That would definitely make this easier to reason about.

mbramson: What do you think about just adding a @type for pair_session that does the same thing? No explicit struct since the PairController is the only place this is used, but still gain a lot of clarity in the type_specs. I'm not sure its worth encoding this structure as a return type, but I'm not sure. Thoughts?

web/controllers/pair_controller.ex
ericworkman: Yeah, I think that is reasonable.

web/controllers/pair_controller.ex
ericworkman: Its also used in the group pair controller iirc. At least a @type would be a good start.

mbramson: It's kind of sort of used in the GroupPairController. There's actually a bit of refactoring that should occur there as well to mirror the fact that technically in the PairController, it's now legal to be in multiple pairs in one week for one group. This results in some buggy behavior in the GroupPairController, which is fine for now since this state can only be reached if something has gone horribly wrong, or someone is poking the database manually.

mbramson:
@ericworkman, I believe that I've addressed your concerns. Take a look!

ericworkman (approved):
Looks great! Awesome job!

PR #148, opened 2017-03-23
Add an active column to the user list on the group show

mbramson:
Just a little additional helpful information. Takes a tiny step towards user specific activity.


ericworkman (approved):
LGTM

PR #150, opened 2017-03-27
Add group specific admin

### LINKED ISSUE #149: Add Group-specific Admins
Add the ability for users to be admins of a specific group. The owner should be able to make users in a group admins of that group. Admins should also be able to make other users admins.

The scope of this card is limited to adding the ability to be an admin and the ability to make other users admins or not.

mbramson:
Adds an is_admin field to the UserGroup schema. Right now this functionality is pretty limited as additional functionality will be added in other cards.

Allows the owner of a group (or other admins) to make members of a group admin's as well.

Adds the edit and update actions to the UserGroupController, which govern modifying a user's membership in a group (currently only able to alter a user's is_admin property).

The edit form also has a link to removing a user from the group, but currently only the owner can do so (not group admins).

Closes #149

### DIFF
--- a/priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
+++ b/priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
@@ -1,0 +1,21 @@
+defmodule :"Elixir.Pairmotron.Repo.Migrations.Add-user-group-is-admin" do
+  use Ecto.Migration
+
+  def up do
+    alter table(:users_groups) do
+      add :is_admin, :boolean
+    end
+
+    execute "UPDATE users_groups SET is_admin = FALSE where is_admin IS NULL"
+
+    alter table (:users_groups) do
+      modify :is_admin, :boolean, null: false, default: false
+    end
+  end
+
+  def down do
+    alter table(:users_groups) do
+      remove :is_admin
+    end
+  end
+end
--- a/priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
+++ b/priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
@@ -1,21 +1,9 @@
 defmodule :"Elixir.Pairmotron.Repo.Migrations.Add-user-group-is-admin" do
   use Ecto.Migration
 
-  def up do
-    alter table(:users_groups) do
-      add :is_admin, :boolean
-    end
-
-    execute "UPDATE users_groups SET is_admin = FALSE where is_admin IS NULL"
-
+  def change do
     alter table (:users_groups) do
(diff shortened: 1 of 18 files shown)


priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
```diff
@@ -0,0 +1,21 @@
...
+
+  def up do
+    alter table(:users_groups) do
+      add :is_admin, :boolean
+    end
+
+    execute "UPDATE users_groups SET is_admin = FALSE where is_admin IS NULL"
+
+    alter table (:users_groups) do
+      modify :is_admin, :boolean, null: false, default: false
```
ericworkman: This is to avoid a full table rewrite when adding the column with a default not-null, right? In this case, since you are adding the column in the same transaction you are setting the value, aren't you also doing a full table rewrite?

web/controllers/user_group_controller.ex
```diff
@@ -10,6 +10,66 @@ defmodule Pairmotron.UserGroupController do
...
+        redirect_not_authorized(conn, group_path(conn, :show, group_id))
+      is_nil(logged_in_users_user_group) ->
+        redirect_not_authorized(conn, group_path(conn, :show, group_id))
+      user_group.group.owner_id == user.id ->
+        render_edit(conn, user_group)
+      logged_in_users_user_group.is_admin ->
+        render_edit(conn, user_group)
+      true ->
+        redirect_not_authorized(conn, group_path(conn, :show, group_id))
+    end
```
ericworkman: 👀 😮 I don't see a cleaner alternative for this.

ericworkman: Should this one have the `user.is_admin` cond?

priv/repo/migrations/20170326201602_add-user-group-is-admin.exs
mbramson: nope, honestly for some reason I thought that this had to be done in two steps because I did a similar (but not the same) thing somewhere else. Derp.

web/controllers/user_group_controller.ex
mbramson: Yep! There's probably a better way to do this utilizing `with`, but that's a thought for a different day I think. I've been thinking of refactoring all controllers to use with and just use a case at the end to prepare for the eventual use of a FallBackController, since that structure would make the coming code migration a lot easier.

mbramson:
@ericworkman I believe I've addressed concerns on this as well, let me know what you think!

ericworkman (approved):
LGTM. Nice!

PR #155, opened 2017-04-02
Replace ex admin

mbramson:
Remove ex admin from the project in favor of standard crud routes. This paves the way to being able to bump phoenix to 1.3, and generally removes a dependency that has given us a number of issues.


ericworkman:
@mbramson This is a lot to review, so it's going to take me some time. A quick glance and pull-and-run looks good to me, but I'd like to review it in more detail. 👍 for removing exadmin

web/controllers/admin/group_membership_request_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+  def new(conn, _params) do
+    changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{})
+    groups = Repo.all(Pairmotron.Group)
+    users = Repo.all(Pairmotron.User)
+    render(conn, "new.html", changeset: changeset, groups: groups, users: users)
+  end
+
+  @spec create(Plug.Conn.t, map()) :: Plug.Conn.t
+  def create(conn, %{"group_membership_request" => group_membership_request_params}) do
+    changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{}, group_membership_request_params)
```
ericworkman: This works, but it would probably be better to use `users_changeset`.

web/controllers/admin/user_controller.ex
```diff
@@ -0,0 +1,72 @@
...
+
+  @spec new(Plug.Conn.t, map()) :: Plug.Conn.t
+  def new(conn, _params) do
+    changeset = User.changeset(%User{})
+    render(conn, "new.html", changeset: changeset)
+  end
+
+  @spec create(Plug.Conn.t, map()) :: Plug.Conn.t
+  def create(conn, %{"user" => user_params}) do
+    changeset = User.changeset(%User{}, user_params)
```
ericworkman, line 20: I think you should use `registration_changeset` so that the password validation happens.

ericworkman (changes requested):
Awesome job! Only these couple of things stick out to me.

web/controllers/admin/user_controller.ex
mbramson: Password validation still happens with the regular changeset. Additionally, using the `registration_changeset` would not allow admins to make other users admins.

web/controllers/admin/group_membership_request_controller.ex
mbramson: Good call!

web/controllers/admin/user_controller.ex
ericworkman: Good call!

web/controllers/admin/group_membership_request_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+  def new(conn, _params) do
+    changeset = GroupMembershipRequest.changeset(%GroupMembershipRequest{})
+    groups = Repo.all(Pairmotron.Group)
+    users = Repo.all(Pairmotron.User)
+    render(conn, "new.html", changeset: changeset, groups: groups, users: users)
+  end
+
+  @spec create(Plug.Conn.t, map()) :: Plug.Conn.t
+  def create(conn, %{"group_membership_request" => group_membership_request_params}) do
+    changeset = GroupMembershipRequest.users_changeset(%GroupMembershipRequest{}, group_membership_request_params)
```
ericworkman: `users_changeset` requires the third argument of `group`.

ericworkman (changes requested):

web/controllers/admin/group_membership_request_controller.ex
mbramson: Doh

mbramson:
Changes made, let me know what you think!

ericworkman (approved):

ericworkman:
Awesome job! Looks good! :shipit:

PR #156, opened 2017-04-03
Allow group admins to edit groups and modify group invitations

### LINKED ISSUE #153: Allow group admins to edit the group
Currently only the owner of a group can edit a group. Allow group admins to also edit their groups. This will involve some additional authorization checks on the group edit and update actions.

Group admins were implemented in #150 and are defined as a user whose UserGroup association with the group they are an admin of has an is_admin property of true.

You will probably want to use the Ecto query defined by `Pairmotron.UserGroup.user_group_for_user_and_group(group_id)` to determine if the current user is the group admin for the current group.

In addition, the UI on the `group/group_actions.html.eex` template needs to be updated so that group admins have a button to click.

Currently a group looks like this for an admin of the group when viewed on the group index or the group show for that group:
screen shot 2017-03-27 at 10 23 37 pm

The following button needs to be visible (it already exists) if the user is a group admin of that group (group owners and global admins should still be able to see it):
screen shot 2017-03-27 at 10 21 26 pm

### LINKED ISSUE #154: Allow group admins to manage invitations
Currently only group owners can manage the invitations for a group. Group admins should be able to view, create, accept, and reject invitations on behalf of the group that they are an admin of. This will involve giving group admins access to all actions of the GroupInvitationsController.

Group admins were implemented in #150 and are defined as a user whose UserGroup association with the group they are an admin of has an is_admin property of true.

You will probably want to use the Ecto query defined by `Pairmotron.UserGroup.user_group_for_user_and_group(group_id)` to determine if the current user is the group admin for the current group.

In addition, group admins should be able to see the "invitations" link on the `group/group_actions.html.eex` template that group owners can see.

This is what they currently see:

screen shot 2017-03-28 at 11 01 39 am

Here's the button that they need to be able to see:
screen shot 2017-03-28 at 11 05 44 am

mbramson:
Gives group admins access to the GroupController :edit and :update actions, as well as the GroupInvitationController :index, :new, :create, :update, and :delete actions.

Makes the "Invitations" and "Edit" links for the group_actions template visible for group admins in addition to group owners.

Closes #153 
Closes #154

### DIFF
--- a/lib/abilities.ex
+++ b/lib/abilities.ex
@@ -1,5 +1,5 @@
 defimpl Canada.Can, for: Pairmotron.User do
-  alias Pairmotron.{Group, PairRetro, Project, Types, User}
+  alias Pairmotron.{Group, PairRetro, Project, Repo, Types, User, UserGroup}
 
   @spec can?(Types.user, atom(), struct()) :: boolean()
   def can?(%User{is_admin: true}, _, _), do: true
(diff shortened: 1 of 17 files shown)


ericworkman (approved):
LGTM

PR #158, opened 2017-05-04
157 - Add Password Reset

### LINKED ISSUE #157: Implement Password Resets
This will likely also involving adding something like bamboo as a dependency to handle sending emails.

mbramson:
Users will be able to reset their passwords via an email reset if they have forgotten their passwords.

TODO:
- [x] Add PasswordResetToken model
- [x] Add /forgotten_password route
- [x] Add form for requesting a password reset email
- [x] Implement PasswordResetToken Service
  - [x] Implement generate_token/1 which takes an email and returns {:ok, token} when that email is assocated with a user.
  - [x] Implement verify_token/2 which takes an email and a token string and verifies that a token exists with that email and token.
- [x] Add `:create` action for controller governing creation of a new password reset token that uses the PasswordResetToken Service.
- [x] Add route for resetting your password to be linked to from the password reset email
- [x] Add actions for resetting password to ForgotPasswordController
  - [x] Add `:edit`action to ForgotPasswordController for users to enter new passwords
    - [x] Make sure token has not expired
  - [x] Add `:update` action for users to change their passwords and delete the used PasswordResetToken
    - [x] Make sure token has not expired
    - [x] Upon success, log the user in
- [x] Add Bamboo dependency
  - [x] Configure Bamboo to use Mailgun
  - [x] Make the forgotten password form send an email containing a link with the token when submitted
- [x] Add forgot password link to login.

Cleanup TODO:
- [x] Use Ecto.Multi in PasswordResetController :update action
- [x] Fix credo issues

Closes #157


lib/mailer/email.ex
```diff
@@ -0,0 +1,48 @@
...
+    |> subject("Pairmotron Password Reset")
+    |> html_body(password_reset_email_html_body(password_reset_token.token))
+    |> text_body(password_reset_email_text_body(password_reset_token.token))
+  end
+
+  @spec password_reset_email_html_body(String.t) :: String.t
+  defp password_reset_email_html_body(token_string) do
+    """
+    It looks like you've requested a password reset.<br><br>
+    <a href=#{reset_path(token_string)}>Click here to reset your password</a>
```
mbramson: Feel free to suggest alternative type here. I kept it pretty simple, but I'm sure some prettier language could be suggested here.

lib/password_reset_token/password_reset_token_service.ex
```diff
@@ -0,0 +1,93 @@
...
+  of the form {:error, :no_user_with_email}.
+
+  generate/token1 also handles retrying in the unlikely case that the randomly
+  generated token is already associated with a PasswordResetToken.
+  """
+  @spec generate_token(String.t) :: {:ok, Types.password_reset_token} | {:error, atom()}
+  def generate_token(email) when is_binary(email) do
+    with {:ok, user} <- get_user_with_email(email),
+         {:ok, token} <- create_token(user),
+      do: {:ok, Repo.preload(token, :user)} # :( for extra database hit
```
mbramson: So I couldn't find a way to do a database insert and have the record returned by Ecto have an association preloaded (thus avoiding a second trip to the database).

This could probably be accomplished with an sql fragment of some sort, but I'm not sure it's worth adding that complexity for such a small optimization.

lib/password_reset_token/password_reset_token_service.ex
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  @spec create_token(Types.user) :: Types.password_reset_token | {:error, :invalid_token}
+  defp create_token(%User{} = user) do
+    token = random_urlsafe_base64()
+    changeset = PasswordResetToken.changeset(%PasswordResetToken{}, %{user_id: user.id, token: token})
+
+    case Repo.insert(changeset) do
+      {:ok, valid_token} -> {:ok, valid_token}
+      {:error, %{errors: [token: {"has already been taken", _}]}} -> create_token(user)
```
mbramson: So I thought about limiting this to some number of attempts to prevent infinite recursion, but I figured the likelihood that this case gets hit twice is so ridiculously low anyhow, that it wouldn't be worth it. I'm not even sure its worth it to have the retry mechanism at all honestly.

The argument for limiting the number of times it can be hit it is basically if the random_urlsafe_base64/0 breaks in some fashion that it just always returns the same thing, it will prevent infinite recursion.

lib/password_reset_token/password_reset_token_service.ex
mbramson: Also sidenote: `with` is so awesome it's ridiculous for cleaning up control flow.

test/controllers/password_reset_controller_test.exs
```diff
@@ -0,0 +1,110 @@
...
+      conn = post conn, password_reset_path(conn, :create), password_reset_token: %{email: user.email}
+      assert Repo.get_by(PasswordResetToken, user_id: user.id)
+      assert html_response(conn, 200) =~ "An email with password reset instructions has been sent"
+    end
+
+    test "sends a password reset email if user with the given email exists", %{conn: conn} do
+      user = insert(:user)
+      post conn, password_reset_path(conn, :create), password_reset_token: %{email: user.email}
+      token = Repo.get_by(PasswordResetToken, user_id: user.id) |> Repo.preload(:user)
+      assert_delivered_email Pairmotron.Email.password_reset_email(token)
```
mbramson, line 26: Courtesy of Bamboo.Test. The Bamboo Adapters are just genservers. That has an awesome effect for testing. You can just swap out whatever adapter is used in dev/prod for a testing adapter in the `test` mix environment. Super handily, Bamboo comes with the `Bamboo.TestAdapter`, which basically just allows you to poke it after it receives an email in your tests.

test/controllers/password_reset_controller_test.exs
```diff
@@ -0,0 +1,115 @@
...
+
+    test "does not create a PasswordResetToken if user does not exist with given email", %{conn: conn} do
+      conn = post conn, password_reset_path(conn, :create), password_reset_token: %{email: "null@email.com"}
+      assert [] = Repo.all(PasswordResetToken)
+      assert html_response(conn, 200) =~ "An email with password reset instructions has been sent"
+    end
+
+    test "does not send an email if user does not exist with the given email", %{conn: conn} do
+      post conn, password_reset_path(conn, :create), password_reset_token: %{email: "null@email.com"}
+      assert_no_emails_delivered()
```
mbramson, line 37: We also get this!

web/controllers/password_reset_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+            render(conn, "reset.html", changeset: changeset, token_string: token_string)
+        end
+      {:error, :token_not_found} ->
+        conn
+        |> put_flash(:error, "Sorry, that is not a valid password reset token")
+        |> redirect(to: session_path(conn, :new))
+      {:error, :token_expired} ->
+        conn
+        |> put_flash(:error, "Sorry, that password reset token has expired.")
+        |> redirect(to: session_path(conn, :new))
```
mbramson: I thought about also deleting the password reset token when this happens. Sort of helps clean up tokens that aren't usable anymore. A better way to do this is probably to set up some sort of weekly job that just deletes expired tokens. Definitely out of the scope of this card.

lib/password_reset_token/password_reset_token_service.ex
```diff
@@ -0,0 +1,93 @@
+defmodule Pairmotron.PasswordResetTokenService do
```
mbramson, line 1: So I put this in it's own lib folder to sort of prepare for the 1.3 directory structure. This file would essentially be the context for the `Pairmotron.PasswordResetToken` schema.

web/models/user.ex
```diff
@@ -87,6 +87,21 @@ defmodule Pairmotron.User do
...
+  new password. Only allows the password to be changed, and also required it to
+  be changed.
+  """
+  @spec password_reset_changeset(map() | %Ecto.Changeset{}, map()) :: %Ecto.Changeset{}
+  def password_reset_changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, @password_reset_params)
+    |> validate_required(@password_reset_params)
+    |> common_changeset
+  end
```
mbramson, line 103: Gotta guard against param injection allowing people to change all sorts of things through the password_reset form.

web/templates/password_reset/reset.html.eex
```diff
@@ -0,0 +1,27 @@
...
+  <div class="form-group">
+    <%= label f, :password_confirmation, class: "control-label" %>
+    <%= password_input f, :password_confirmation, class: "form-control" %>
+    <%= error_tag f, :password_confirmation %>
+  </div>
+
+  <div class="form-group">
+    <%= submit "Submit", class: "btn btn-primary pull-right" %>
+  </div>
+<% end %>
```
mbramson, line 27: So I thought about forcing the user to re-enter the email associated with the user, but I thought it was overkill. If the token is intercepted in some way, it's pretty likely that the attacker already has the email associated with that token. I can't think of a situation where this wouldn't be the case.

That said, doing so might be a good guard against this code malfunctioning in the future.

I think keeping it simple and leaving off the email validation is probably the correct thing to do.

web/templates/session/new.html.eex
```diff
@@ -15,4 +15,8 @@
     <%= link("Sign Up", to: registration_path(@conn, :new), class: "btn btn-success") %>
     <%= submit "Login", class: "btn btn-primary pull-right" %>
 </div>
+
+<p>
+  <%= link("Forgot your password", to: password_reset_path(@conn, :new)) %>
```
mbramson: Totally up to some different formatting here or alternative language. Thoughts?

mbramson (commented):
Thoughts and commentary.

lib/mailer/email.ex
```diff
@@ -0,0 +1,48 @@
...
+    """
+  end
+
+  @spec password_reset_email_text_body(String.t) :: String.t
+  defp password_reset_email_text_body(token_string) do
+    """
+    It looks like you've requested a passwrod reset.\n\n
```
ericworkman: `passwrod`

lib/password_reset_token/password_reset_token_service.ex
ericworkman: Neato

lib/password_reset_token/password_reset_token_service.ex
ericworkman: `with` is definitely great.

ericworkman: As far as I know, you are right. You can't do a preloaded returning insert. You could move the preload into `create_token/1` if it makes more sense there.

lib/password_reset_token/password_reset_token_service.ex
ericworkman: I think we'd have bigger problems if `:crypto.strong_rand_bytes/1` gave a duplicate byte sequence. I don't think there's a chance that `Base.url_encode64/1` could return the same encoded string for two different binary sequences.

That said, can you think of anyway this could be abused? I'm leaning towards removing the retry just to make it simple.

test/controllers/password_reset_controller_test.exs
ericworkman: Nice 👍

web/controllers/password_reset_controller.ex
ericworkman: It might be better to use the same "not a valid password reset token" instead of specifically saying its expired. Give those looking to exploit it less information.

web/models/user.ex
ericworkman: 👍

web/templates/password_reset/email_sent.html.eex
```diff
@@ -0,0 +1,4 @@
+<h3>Forgot Password</h3>
+<p class="lead">
+  An email with password reset instructions has been sent if that email is associated with a user.
```
ericworkman: I'd just leave it to `An email with password reset instructions has been sent.`

web/templates/password_reset/reset.html.eex
ericworkman: Yep, that really won't help if the attacker knows the email.

web/templates/profile/show.html.eex
```diff
@@ -23,8 +23,8 @@
     </div>
   </div>
   <div class="col-md-7 col-sm-12">
-    <h3>Member of</h3>
     <%= for group <- @user.groups do %>
+      <h3>Member of</h3>
```
ericworkman: So now we show a `Member of` for each group? Does that look weird?

web/templates/session/new.html.eex
ericworkman: I'd just add a ? to that.

README.md
```diff
@@ -26,7 +26,13 @@ export PG_PASSWORD="example_password"
...
+## Ready to run in production?
 * Please [check the phoenix deployment guides](http://www.phoenixframework.org/docs/deployment).
-* Set the environment variable (SECRET_KEY_BASE) used for session cookies.
-* Set the environment variable (GUARDIAN_JWK_KEY) for the secret key which guardian uses for JWT tokens.
+
+### Environment Variables to set
+
+* SECRET_KEY_BASE - Used to generate session cookies.
+* GUARDIAN_JWK_KEY - Used to generate JWT tokens.
+* PAIRMOTRON_EMAIL_DOMAIN - The domain configured in Mailgun to send password reset email.
```
ericworkman, line 36: Are these new env vars required to run pairmotron in dev, or can we continue running locally this without them?

lib/mailer/email.ex
ericworkman: Maybe something like

```
Someone requested a password change for your Pairmotron account. If this was requested by you, please set a new password here:

<password reset link>

If you didn't request a password change...
```

I'd hate to have a contact us at the end there, so we could just say "ignore this email" or something, but that rubs me the wrong way.

lib/mailer/email.ex
```diff
@@ -0,0 +1,48 @@
...
+  given password_reset_token. The email contains a link to the
+  PasswordResetController's :edit form where the user can set a new password.
+
+  The password_reset_token's :user relation must be preloaded.
+  """
+  @spec password_reset_email(Types.password_reset_token) :: Bamboo.Email.t
+  def password_reset_email(password_reset_token)  do
+    new_email()
+    |> to(password_reset_token.user.email)
+    |> from("no-reply@pairmotron.com")
```
ericworkman, line 22: You don't own pairmotron.com, do you?

ericworkman (changes requested):

lib/password_reset_token/password_reset_token_service.ex
mbramson: Ya that's a good point. The chance of a collision is so ridiculously low, and it really is just sort of opening up a subtle attack vector.

web/controllers/password_reset_controller.ex
mbramson: Good call

lib/mailer/email.ex
mbramson: I do!

web/templates/profile/show.html.eex
mbramson: D'oh. Ya I was testing on a single group, this didn't cross my mind. Good catch.

README.md
mbramson: they are not required, since `System.get_env/1` returns nil if the config isn't found. Which isn't actually used by Bamboo until it tries to send an email (at which point it obviously fails). So it really only fails if you try to request a password_reset and it isn't properly configured.

lib/password_reset_token/password_reset_token_service.ex
mbramson: Ya I like it in create_token better. It keeps the operations with the database in one place.

lib/mailer/email.ex
mbramson: I like that better. I'm going to leave off the "If you didn't request a password change..." bit as I'm not going to be looking at any email on the pairmotron.com domain anyhow. Something to keep in mind going forward for sure.

mbramson:
@ericworkman I believe that I've addressed your concerns.

ericworkman:
LGTM. The only request I have is that we get pairmotron.com pointing to the project so that all the email verification address are a-okay.

ericworkman (approved):

ericworkman:
Hold off merging until we get the dns changes in, please 😃

PR #159, opened 2017-06-03
Make production use ES512 instead of HS512

mbramson:
This fixes a bit of technical debt from back in the day when we first added guardian tokens. We got ES512 to work in dev, but couldn't quite get it to work in prod, so we just used HS512.

This PR makes production use ES512 rather than HS512. ES512 is recommended over HS512 when used with the P-521 curve based on this: https://github.com/ueberauth/guardian/issues/152.


ericworkman (approved):
LTGM! Let's get these env vars in and merge this in.

PR #164, opened 2017-06-30
Add email for when a user is invited to a group

### LINKED ISSUE #163: Email users when they are invited to groups
Currently when a user is invited to a group, this happens essentially silently. They either have to go to their invitations explicitly, or if #162 has happened, be in the app and see a notification.

It would be great if we emailed users when this type of thing happened, preferably with a link to immediately accept the invitation (linking to the `UserGroupMembershipRequestController`'s `:update` route if possible.

mbramson:
Send an email to users who have been invited to a group by the group's owner. This email informs them that they have been invited to the group and links them to the listing of their current invitations.

Closes #163


ericworkman (approved):
Looks excellent to me!

PR #166, opened 2017-06-30
Improve the Admin UI

mbramson:
- Alphabetize various dropdowns on the admin UI
- Change the format of entries in admin dropdowns from "id: name" to "name (id)" for easier searching with keyboard.

### DIFF
--- a/web/controllers/admin/user_group_controller.ex
+++ b/web/controllers/admin/user_group_controller.ex
@@ -76,10 +76,10 @@ defmodule Pairmotron.AdminUserGroupController do
 
   @spec retrieve_groups_and_users() :: {Types.group, Types.user}
   defp retrieve_groups_and_users() do
-    groups = Group 
+    groups = Group
       |> Repo.all
       |> Enum.sort(&(String.downcase(&1.name) <= String.downcase(&2.name)))
-    users = User 
+    users = User
       |> Repo.all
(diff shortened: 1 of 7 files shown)


web/controllers/admin/user_group_controller.ex
```diff
@@ -43,17 +41,16 @@ defmodule Pairmotron.AdminUserGroupController do
...
   end
 
   @spec update(Plug.Conn.t, map()) :: Plug.Conn.t
   def update(conn, %{"id" => id, "user_group" => user_group_params}) do
     user_group = Repo.get!(UserGroup, id)
     changeset = UserGroup.changeset(user_group, user_group_params)
-    groups = Repo.all(Pairmotron.Group)
-    users = Repo.all(Pairmotron.User)
+    {groups, users} = retrieve_groups_and_users()
+    render(conn, "edit.html", user_group: user_group, changeset: changeset, groups: groups, users: users)
```
ericworkman: I don't think this is supposed to be here, right?

ericworkman (changes requested):
Just the one line

web/controllers/admin/user_group_controller.ex
mbramson: nope! good catch

ericworkman (approved):

PR #170, opened 2017-07-03
Remove ex_admin images, add bg image, add some css fixes

ericworkman:
@mbramson I liked what @Serneum did with tabletop, so I did something similar. 

image


mbramson (approved):
LGTM

Serneum (approved):
LGTM

Serneum:
@ericworkman I forgot to actually have you check what happens if the container is longer than the view window. In my case, I had to add `height: 100%` to `body` so that it would render properly

ericworkman:
@Serneum Have a screenshot of that?

Serneum:
I will try and reproduce it, but I can describe a bit of what I was seeing. The list of Games in CoverMyTableTop was longer than the what was currently visible in the browser, so once you scrolled down, the container would just stop and the list would keep going beyond the bounds of the container. Instead of having the white background, they'd be on top of the background image

I had solved it by adding a `min-height: 100%`, not a `height: 100%`

ericworkman:
I see. Yeah, please try to reproduce it. So far I'm not able to.

PR #161, opened 2017-06-25
Add docker support for development and testing

devertap:
In this pull request, I've added a base elixir image with some minor dependencies.  Then it is wired up using docker-compose in conjunction with a standard postgres:9.5 image.  

I then documented it via the README.  You'll notice I changed the config for the local PG_USER/PASSWORD.  This was to make them the same and accessible via the dev.exs and test.exs.

Hope it helps.

PR #171, opened 2017-07-05
Adding docker support for development and test environments. 

devertap:
This PR adds docker support for both the development and testing environments.  Also, setting travis and readme instructions to help facilitate the move.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ otp_release:
 addons:
   - postgresql: '9.5'
 env:
-  - MIX_ENV=test
-  - PG_USER=postgres
+  - MIX_ENV=test PG_USER=postgres
 install:
   - mix local.hex --force
   - mix local.rebar --force
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -39,5 +39,6 @@ config :pairmotron, Pairmotron.Repo,
   username: System.get_env("PG_USER"),
   password: System.get_env("PG_PASSWORD"),
   database: "pairmotron_dev",
-  hostname: "localhost",
+  hostname: System.get_env("PG_HOST") || "localhost",
   pool_size: 10
+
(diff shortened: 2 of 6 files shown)


ericworkman (approved):
This is just a clean up of @devertap 's #161 . LGTM

config/dev.exs
```diff
@@ -39,5 +39,6 @@ config :pairmotron, Pairmotron.Repo,
   username: System.get_env("PG_USER"),
   password: System.get_env("PG_PASSWORD"),
   database: "pairmotron_dev",
-  hostname: "localhost",
+  hostname: System.get_env("PG_HOST") || "localhost",
```
Serneum, line 42: Could this be set in `config/config.exs` as a default?

Serneum: Nevermind, that might break prod

Serneum (approved):
LGTM

config/dev.exs
devertap: Yeah, that was my concern and why I chose to leave it here at this level.

.travis.yml
```diff
@@ -6,8 +6,7 @@ otp_release:
 addons:
   - postgresql: '9.5'
 env:
-  - MIX_ENV=test
-  - PG_USER=postgres
+  - MIX_ENV=test PG_USER=postgres
```
devertap, line 9: Also, this line might have been better suited in a separate PR, but this changes the travis job from running two steps and combines the config it into a single job.  When you review details, since there is only one thing to do it shows the output automatically.  Kind of a nice win.

mbramson: Yep, this is great!

mbramson (approved):
LGTM

ericworkman:
Thanks for the PR @devertap !

PR #174, opened 2017-07-10
Fix error when updating retro with no project

mbramson:
Fixes an error when updating a retro that has no associated project.

Looks like there was no test coverage around this case for update, while there was for the create. Uncovered this trying to change the date of a retro that didn't have an associated project in production.


ericworkman (approved):
LGTM

PR #175, opened 2017-11-13
Add accomplice for grouping

mbramson:
Adds the Accomplice library which can pair up any number of users to replace the current logic in the `Pairer` module. This offloads a bunch of that logic to a library that is more generic, and generally make a lot of the code in the `Pairer` module a lot easier to understand.

This also paves the way for future groups to be able to configure what sizes of groups they'd like to generate, as the Accomplice library allows the specification of these parameters.

This also paves the way for groups to be able to prevent users from previous weeks from pairing again, as well as giving users the ability to specify that they would like to be paired with someone specifically the following week.

This also adds the OrderInvariantCompare library which can be useful for testing lists lists of values when you don't really care about the order of the elements of the returned lists, you just care about how many elements are in each list.

### DIFF
--- a/lib/pairmotron/pairer.ex
+++ b/lib/pairmotron/pairer.ex
@@ -25,23 +25,20 @@ defmodule Pairmotron.Pairer do
   end
 
   @spec generate_pairs([Types.user]) :: %PairerResult{pairs: [Types.pair]}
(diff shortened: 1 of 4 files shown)


ericworkman (approved):
Looks great!

